PR #331, opened 2014-03-13
Lazy split

gsigms:
Refactor lazy evaluation:
* change granularity of lazy to halo send/recv, compute core/owned/halo, and reduction begin/end instead of whole ```ParLoop``` computation.
* maintain delayed computations in a directed acyclic graph instead of a linear trace.
* schedule halo send for early execution when possible when planning execution of delayed computations


kynan:
Is this still something we want to pursue?

dham:
too old

PR #347, opened 2014-04-07
IR is not reproducibly hashable

wence-:
If we are passed a kernel as an AST, we use its hash as part of the cache key to look up whether we still have it.  This feeds into the JITModule cache key which tells us whether we already have a compiled function pointer in memory.

Constructing the same AST twice, should produce an identical kernel cache key.  However, it does not, because the hash of the AST is not the same in both cases.

One notices this particularly in firedrake now that Expression assembly uses the AST (we don't hit the cache as often as we should).  Is it enough to just cache based on the generated code?  Will that cause false positives if we change the optimisation settings?

### DIFF
--- a/pyop2/ir/ast_base.py
+++ b/pyop2/ir/ast_base.py
@@ -66,7 +66,16 @@ class Node(object):
     """The base class of the AST."""
 
     def __init__(self, children=None):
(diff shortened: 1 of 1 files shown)


wence-:
I think this is currently not a problem, closing.

PR #402, opened 2014-09-22
Baij matrices

wence-:
Here's the more controversial part.  Block-sparse matrices.


kynan:
Has this been superseded by #431?

wence-:
Oh, yes.

PR #404, opened 2014-09-25
Reduce overhead

wence-:
A bunch of changes to reduce the overhead of calling the same par_loop again and again.

Summary:

- purges lazy evaluation (since I have not thought hard enough about how the new features should interact and we currently don't use it anyway).

- reduces the amount of type-checking ctypes does when calling a compiled function: we generate the code so the types are "guaranteed" to be correct.

- Exposes the ParLoop object in the public API
  You can now repeatedly run the same par_loop (with the same arguments) by doing:
```python
   pl = ParLoop(.....)
   for i in range(100):
       pl._run()
```

- adds a (no-error-checking) `replace_arg_data` routine that replaces the data payload of a specified argument in a ParLoop.  That means you can cons up a new Dat or Mat, replace it in a saved ParLoop and call the ParLoop again writing/reading the new (rather than original) object.

- various minor adjustments to kill unnecessary repeated computations when rerunning a parloop.

I think the only truly controversial part of this is the lazy evaluation purge.  Comments welcome.


kynan:
This has gone stale. Is there a sensible way to land the changes not purging lazy? Or is that no longer controversial?

FabioLuporini:
My work now totally relies on lazy; so if we land this, lazy should stay

wence-:
I have various ideas about how to do this right.  But it needs quite a bit of rewiring and so forth.  But this can be dropped, I think.  I'll leave the branch around as part of the ideas.

kynan:
@wence- Has this progressed since?

miklos1:
@kynan: Of course, OP2/PyOP2#435 and firedrakeproject/firedrake#500.

PR #419, opened 2015-01-05
Remove bad instruction from CONTRIBUTING.md

chromy:
It doesn't make sense to overwrite the thing we just downloaded with
the sample hook.


kynan:
Indeed not, thanks!

PR #420, opened 2015-01-05
Fix skipping CUDA for TestDat and TestMatrices

chromy:
The idiom for skipping cuda seems to have changed but these two places still
use the old way which now crashes when you run with the CUDA backend.

The crash looked like this:
```
Error evaluating 'xfail' expression
    config.getvalue("backend")[0] == "cuda"
TypeError: 'NoneType' object has no attribute '__getitem__'
```

I couldn't work out exactly why this was happening but I think the rest of the tests
now use the skip_x fixtures so we can use them instead.

### DIFF
--- a/test/unit/test_dats.py
+++ b/test/unit/test_dats.py
@@ -108,7 +108,7 @@ class TestDat:
         with pytest.raises(NotImplementedError):
             mdat.copy(op2.MixedDat([s, s]), subset=op2.Subset(s, []))
 
-    @pytest.mark.skipif('config.getvalue("backend")[0] not in ["cuda", "opencl"]')
+    @pytest.mark.skipif('config.getvalue("backend") and config.getvalue("backend")[0] not in ["cuda", "opencl"]')
     def test_copy_works_device_to_device(self, backend, d1):
         d2 = op2.Dat(d1)
 
--- a/test/unit/test_matrices.py
+++ b/test/unit/test_matrices.py
@@ -762,7 +762,7 @@ class TestMatrices:
         """Check that the matrix uses the amount of memory we expect."""
         assert mat.nbytes == 14 * 8
 
-    @pytest.mark.xfail('config.getvalue("backend")[0] == "cuda"')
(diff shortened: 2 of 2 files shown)


test/unit/test_dats.py
```diff
@@ -108,8 +108,7 @@ def test_copy_mixed_subset_fails(self, backend, s, mdat):
         with pytest.raises(NotImplementedError):
             mdat.copy(op2.MixedDat([s, s]), subset=op2.Subset(s, []))
 
-    @pytest.mark.skipif('config.getvalue("backend")[0] not in ["cuda", "opencl"]')
-    def test_copy_works_device_to_device(self, backend, d1):
+    def test_copy_works_device_to_device(self, backend, d1, skip_cuda, skip_opencl):
```
kynan: This is the other way round! This test should only run if the backend *is* CUDA or OpenCL.

kynan:
Which version of `py.test` are you using? Maybe the interface for marks has changed.

chromy:
I'm using py.test 2.6.4 but I think I might just be confused about how to run the tests.

I expected to able to run the tests on the cuda backend like this this: `PYOP2_BACKEND=cuda py.test test/unit/test_dats.py`
But I think I should have been doing this:  `py.test --backend=cuda test/unit/test_dats.py`

If  `PYOP2_BACKEND=cuda py.test test/unit/test_dats.py` isn't expected to work then we can probably close this issue.

kynan:
We might be able to make this work if desired (environment variables could be treated the same way as command line arguments).

chromy:
In retrospect I think I was wrong to expect it work, you should probably have to explicitly state what backend you're trying to test. I think the thing that really confused me (other than failing to read the documentation...) was that running with no backend skipped almost all the interesting tests but failed on those two with `config.getvalue("backend")[0]` in the condition so I think that changing those to skip if backend=None would probably help.

kynan:
Yes, I think that's good. @wence-?

kynan:
This has slipped through the cracks: is this good to merge?

PR #421, opened 2015-01-07
The CUDA Mat should implement duplicate

chromy:
The CUDA implmentation of Mat does not seem to end
up with a definition of duplicate. It looks like
the OpenCL gets its from petsc_base.Mat so we should
probably be getting ours from there to.

### DIFF
--- a/pyop2/cuda.py
+++ b/pyop2/cuda.py
@@ -366,6 +366,12 @@ class Mat(DeviceDataMixin, op2.Mat):
         self._lmadata.fill(0)
         self._version_set_zero()
 
+    def duplicate(self):
+        other = Mat(self.sparsity)
+        base._trace.evaluate(set([self]), set([self]))
+        setattr(other, '__csrdata', self._csrdata.copy())
+        return other
+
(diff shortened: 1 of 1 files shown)


wence-:
NAK, the CUDA mat doesn't have a petsc data structure underneath it.  It needs to implement its own duplicate method.  Effectively you just need to build a new Cuda Mat and copy the data array from one to the other.

wence-:
Sparsities shouldn't have to be duplicated.  I would do something like:

```python

def duplicate(self):
    other = Mat(self.sparsity)
    # ensure outstanding par loops have been carried out
    base._trace.evaluate(set([self]), set([self]))
    # or whatever the appropriate pycuda call is
    self._csrdata.copy(other._csrdata)
```

chromy:
hmm, I think I'm a bit out of my depth here :) I'll probably wait until next week when I'm back at Imperial and can talk to people about this before working on this anymore, thanks for your help!

kynan:
Has there been any more work on this?

PR #423, opened 2015-02-06
travis: Install requirements one by one

wence-:
Recent pip updates no longer install packages in requirements.txt in the
order they're specified.  But some of the packages we use depend on
previously having installed other requirements.  Fudge around this
particular Python packaging disaster by passing each line of the
requirements file individually to pip via xargs.  This does mean the
only things we can put in requirements files are package requirements,
oh well.


wence-:
This merge is a disgusting hack to get round updates to pip which means requirements aren't installed in order, ARGH!

PR #424, opened 2015-02-10
Halos callback

wence-:
Rework halos so that the exchange functionality lives on the Halo, not the Dat.  This way a user can provide a different halo exchange routine that just implements the necessary pieces.


kynan:
Looks good to me!

PR #426, opened 2015-03-02
Associate field ISes with DataSet (and MixedDataSet)

wence-:
Rather than consing up a new IS for each Mat we build, save them on the
DataSet, where they belong.


wence-:
Now also stashes the LGMap on the DataSet (again a memory saving of one per dataset, rather than one per Mat mapping to/from the dataset).

PR #428, opened 2015-03-05
make sure is_top is false for non-extruded maps

dham:


wence-:
I guess this is the minimal change to get things to work for firedrakeproject/firedrake#479

dham:
Basically, yes.I think it's also the Right Thing because we're directly checking if the relevant map is extruded.

kynan:
Seems sensible.

kynan:
Is this good to merge @wence-?

PR #431, opened 2015-03-06
Monolithic matrices

wence-:
Add support for assembling into a single monolithic matrix, as opposed to a MatNest block.  This option is selected if the sparsity was built with nest=False.  The default is controlled by the configuration option "matnest", with environment variable PYOP2_MATNEST.  Defaults to True (i.e. MatNest blocks).

Requires this petsc change https://bitbucket.org/petsc/petsc/pull-request/281/unbreak-matcreatelocalref-in-light-of


wence-:
Oh, in addition, this switches to BAIJ matrices for "vector function spaces" on CPU backends.  I've also ripped out all the OpenCL matrix support, because it was always a horrible hack.

pyop2/petsc_base.py
```diff
@@ -98,10 +99,10 @@ def lgmap(self):
 
     @property
     def field_ises(self):
-        """A list of PETSc ISes defining the indices for each set in
-    the DataSet.
+        """A list of PETSc ISes defining the global indices for each set in
+        the DataSet.
 
-        Used when creating block matrices."""
+        Used when extract blocks from matrices for solvers."""
```
kynan: extracting

pyop2/petsc_base.py
```diff
@@ -257,6 +348,114 @@ def vec_ro(self):
...
+
+    def __getitem__(self, idx):
+        return self
+
+    def __iter__(self):
+        yield self
+
+    def inc_local_diagonal_entries(self, rows, diag_val=1.0):
+        rbs, _ = self.dims[0][0]
+        # No need to set anything if we didn't get any rows.
```
kynan: Suppose this could be checked first.

pyop2/petsc_base.py
```diff
@@ -257,6 +348,114 @@ def vec_ro(self):
...
+        pass
+
+    @property
+    def values(self):
+        rset, cset = self._parent.sparsity.dsets
+        rowis = rset.field_ises[self._i]
+        colis = cset.field_ises[self._j]
+        mat = self._parent.handle.getSubMatrix(isrow=rowis,
+                                               iscol=colis)
+        return mat[:, :]
```
kynan: This returns a view rather than a copy?

pyop2/petsc_base.py
```diff
@@ -319,7 +553,8 @@ def _init_block(self):
         mat.setOption(mat.Option.NEW_NONZERO_ALLOCATION_ERR, True)
         # Do not ignore zeros while we fill the initial matrix so that
         # petsc doesn't compress things out.
-        mat.setOption(mat.Option.IGNORE_ZERO_ENTRIES, False)
+        if not block_sparse:
```
kynan: Why is this only needed for non-block sparsities?

kynan:
That is quite a mouthful! It would be easier to review the blocked and monolithic changes separately, but I guess it makes sense to land them together.

I only had the chance to scan through it (in particular I can't claim to understand the numbering magic), so any comments are rather superficial. Looks sane overall though!

pyop2/petsc_base.py
wence-: the mat is a petsc4py mat, indexing it returns a numpy array of the contents (ala the other values property).

pyop2/petsc_base.py
wence-: it's an error for block matrices.

pyop2/base.py
```diff
@@ -3339,6 +3369,12 @@ def ncols(self):
         return self._ncols
 
     @property
+    def nested(self):
+        """Whether a sparsity is monolithic (even if it has a block structure).
```
dham: This comment is a little cryptic. Maybe make it longer and more explicit.

pyop2/base.py
```diff
@@ -3499,6 +3535,26 @@ def dims(self):
         return self._sparsity._dims
 
     @property
+    def nrows(self):
+        "The number of rows in the matrix (local to this process)"
+        return sum(d.size * d.cdim for d in self.sparsity.dsets[0])
+
+    @property
+    def nblock_rows(self):
```
dham: Please to be adding docstrings.

pyop2/base.py
```diff
@@ -3499,6 +3535,26 @@ def dims(self):
...
+        assert len(self.sparsity.dsets[0]) == 1, "Block rows don't make sense for mixed Mats"
+        return self.sparsity.dsets[0].size
+
+    @property
+    def nblock_cols(self):
```
dham: Ditto

pyop2/host.py
```diff
@@ -248,60 +250,74 @@ def c_addto_scalar_field(self, i, j, buf_name, extruded=None, is_facet=False):
...
+                   }
+                }""" % {'nrows': nrows,
+                        'ncols': ncols,
+                        'rbs': rbs,
+                        'cbs': cbs,
+                        'idx_l': idx_l,
+                        'idx_r': idx_r,
+                        'buf_name': buf_name,
+                        'tmp_name': tmp_name}]
+                addto_name = tmp_name
```
dham: This code generation strategy is not  a thing of beauty. But that's a more general host.py problem to be fixed some other time.

dham:
OK. Aside from the trivial commenting issues, I think we should man up and land this. I've read it, Florian's read it, and it's not obviously wrong. It's too complex to really reason out the correctness of the code in wetware but it passes tests, so I think we should land it and pray.

wence-:
OK, the petsc changes are not in yet, and I don't know any way of actually persuading them to merge things that fix demonstrable bugs, do we just merge the petsc side of things into the firedrake petsc branch and hope that it eventually lands in master (at which point we can overwrite).

wence-:
OK, I will merge the relevant petsc changes into the firedrake branch (also for SF halos) and we can land these changes.

wence-:
Have addressed comments and have merged the petsc fix into the firedrake branch.  Will merge pending travis go-ahead.

PR #432, opened 2015-03-31
Loop Fusion

FabioLuporini:
This PR introduces loop fusion in PyOP2. 

Bad news:
- It depends on three other branches (in FFC, COFFEE, and Firedrake)
- It obviously passes all tests, including Firedrake ones (with fusion enabled), but of course this should be tested extensively in bigger codes
- It's quite massive...

Good news:
- ...but the changes are mostly confined in a new module, fusion.py and...
- ...fusion is off by default, but one can switch it on by simply exporting PYOP2_LOOP_FUSION=1. I think this is the best scheme for the moment, so incrementally we can fix bugs and add functionalities. This also ensures no performance regression (if any).
- performance evaluation will be the next step
- anyway, I haven't experienced any tangible overhead due to fusing loops

Other minor points:
- extrusion and mixed are currently not supported in "hard fusion mode" (see below). It's a matter of fixing code generation, which I'll do after some performance experimentation

How it works (summary):
- it's all based on lazy evaluation of loops. If fusion is ON, the trace of loops is intercepted prior to execution and analysed
- three sort of fusions are possible, depending on the actual data dependency pattern between loops: 
  * if no dependency at all or only "trivial" ones are present (e.g., two *direct* loops, write-after-read or read-after-writ), then ``soft fusion`` is performed: that's basically body concatenation, plus a few minor tweaks (e.g. avoiding passing in the same Arg twice, etc.). 
  * if dependencies are less trivial, like increment-after-increment in indirect loops (the case of cell/facet integrals in Firedrake), then what I called ``hard fusion`` takes place. Things here get more complicated, because basically one needs to track which elements of the fused loop (e.g. cells) have already been visited to avoid executing the same iteration multiple times. This is not the only issue though -- however, the code is fairly commented, so if interested I invite to have a look at it.
  * if dependencies are "real dependencies", like read-after-write between indirect loops. then fusion is implemented through ``split tiling``. For this, an ``inspector-executor scheme`` is implemented exploiting the SLOPE library (accessible at coneoproject/SLOPE).  This is an under-development feature: it currently works if one explicitly enabled it,  but still far we are from actually being able of extensively using it (for example, it lacks MPI support).

Details on the implementation:
- Polymorphism is exploited extensively
- Fusion is heavily based on caching. The fused kernels, the data dependency analysis etc, are all performed just once for identical sequences of loops

Required branches/PR:
- PR 39 in coneoproject/COFFEE
-  PR 492 in firedrakeproject/firedrake
- PR 72 in bitbucket.org/mapdes/ffc

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3721,7 +3721,6 @@ class Kernel(Cached):
         self._include_dirs = include_dirs
         self._headers = headers
         self._user_code = user_code
-        self._code = code
         # If an AST is provided, code generation is deferred
         self._ast, self._code = (code, None) if isinstance(code, Node) else (None, code)
         self._initialized = True
@@ -3731,7 +3730,6 @@ class Kernel(Cached):
         """Kernel name, must match the kernel function name in the code."""
         return self._name
 
-    @property
     def code(self):
         """String containing the c code for this kernel routine. This
         code must conform to the OP2 user kernel API."""
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -64,9 +64,10 @@ class LazyComputation(object):
     """Helper class holding computation to be carried later on.
     """
 
-    def __init__(self, reads, writes):
+    def __init__(self, reads, writes, incs):
         self.reads = set(flatten(reads))
         self.writes = set(flatten(writes))
+        self.incs = set(flatten(incs))
         self._scheduled = False
 
     def enqueue(self):
@@ -3513,7 +3514,7 @@ class Mat(SetAssociated):
 
         Called lazily after user calls :meth:`assemble`"""
         def __init__(self, mat):
-            super(Mat._Assembly, self).__init__(reads=mat, writes=mat)
+            super(Mat._Assembly, self).__init__(reads=mat, writes=mat, incs=mat)
             self._mat = mat
 
         def _run(self):
@@ -3868,7 +3869,8 @@ class ParLoop(LazyComputation):
     def __init__(self, kernel, iterset, *args, **kwargs):
         LazyComputation.__init__(self,
                                  set([a.data for a in args if a.access in [READ, RW]]) | Const._defs,
-                                 set([a.data for a in args if a.access in [RW, WRITE, MIN, MAX, INC]]))
+                                 set([a.data for a in args if a.access in [RW, WRITE, MIN, MAX, INC]]),
+                                 set([a.data for a in args if a.access in [INC]]))
         # INCs into globals need to start with zero and then sum back
         # into the input global at the end.  This has the same number
         # of reductions but means that successive par_loops
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3923,7 +3923,7 @@ class ParLoop(LazyComputation):
 
         self._it_space = self.build_itspace(iterset)
 
-        # Attach semantical information to the kernel's AST
+        # Attach semantic information to the kernel's AST
         if hasattr(self._kernel, '_ast') and self._kernel._ast:
             ast_info = ast_visit(self._kernel._ast, search=ast.FunDecl)
             fundecl = ast_info['search'][ast.FunDecl]
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3930,7 +3930,7 @@ class ParLoop(LazyComputation):
             if len(fundecl) == 1:
                 for arg, f_arg in zip(self._actual_args, fundecl[0].args):
                     if arg._uses_itspace and arg._is_INC:
-                        f_arg.pragma = ast.WRITE
+                        f_arg.pragma = set([ast.WRITE])
 
     def _run(self):
         return self.compute()
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -1930,9 +1930,9 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         """Create the :class:`ParLoop` implementing copy."""
         if not hasattr(self, '_copy_kernel'):
             k = ast.FunDecl("void", "copy",
-                            [ast.Decl(self.ctype, ast.Symbol("*self"),
+                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self"),
                                       qualifiers=["const"]),
-                             ast.Decl(other.ctype, ast.Symbol("*other"))],
+                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"))],
                             body=ast.c_for("n", self.cdim,
                                            ast.Assign(ast.Symbol("other", ("n", )),
                                                       ast.Symbol("self", ("n", ))),
@@ -2029,9 +2029,9 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         if np.isscalar(other):
             other = _make_object('Global', 1, data=other)
             k = ast.FunDecl("void", name,
-                            [ast.Decl(self.ctype, ast.Symbol("*self"),
+                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self"),
                                       qualifiers=["const"]),
-                             ast.Decl(other.ctype, ast.Symbol("*other"),
+                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
                                       qualifiers=["const"]),
                              ast.Decl(self.ctype, ast.Symbol("*ret"))],
                             ast.c_for("n", self.cdim,
@@ -2045,9 +2045,9 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         else:
             self._check_shape(other)
             k = ast.FunDecl("void", name,
-                            [ast.Decl(self.ctype, ast.Symbol("*self"),
+                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self"),
                                       qualifiers=["const"]),
-                             ast.Decl(other.ctype, ast.Symbol("*other"),
+                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
                                       qualifiers=["const"]),
                              ast.Decl(self.ctype, ast.Symbol("*ret"))],
                             ast.c_for("n", self.cdim,
@@ -2071,8 +2071,8 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         if np.isscalar(other):
             other = _make_object('Global', 1, data=other)
             k = ast.FunDecl("void", name,
-                            [ast.Decl(self.ctype, ast.Symbol("*self")),
-                             ast.Decl(other.ctype, ast.Symbol("*other"),
+                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self")),
+                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
                                       qualifiers=["const"])],
                             ast.c_for("n", self.cdim,
                                       ops[op](ast.Symbol("self", ("n", )),
@@ -2083,8 +2083,8 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
             self._check_shape(other)
             quals = ["const"] if self is not other else []
             k = ast.FunDecl("void", name,
-                            [ast.Decl(self.ctype, ast.Symbol("*self")),
-                             ast.Decl(other.ctype, ast.Symbol("*other"),
+                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self")),
+                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
                                       qualifiers=quals)],
                             ast.c_for("n", self.cdim,
                                       ops[op](ast.Symbol("self", ("n", )),
@@ -2098,7 +2098,7 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         ops = {operator.sub: ast.Neg}
         name = "uop_%s" % op.__name__
         k = ast.FunDecl("void", name,
-                        [ast.Decl(self.ctype, ast.Symbol("*self"))],
+                        [ast.Decl("%s*" % self.ctype, ast.Symbol("self"))],
                         ast.c_for("n", self.cdim,
                                   ast.Assign(ast.Symbol("self", ("n", )),
                                              ops[op](ast.Symbol("self", ("n", )))),
@@ -2118,9 +2118,9 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         ret = _make_object('Global', 1, data=0, dtype=self.dtype)
 
         k = ast.FunDecl("void", "inner",
-                        [ast.Decl(self.ctype, ast.Symbol("*self"),
+                        [ast.Decl("%s*" % self.ctype, ast.Symbol("self"),
                                   qualifiers=["const"]),
-                         ast.Decl(other.ctype, ast.Symbol("*other"),
+                         ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
                                   qualifiers=["const"]),
                          ast.Decl(self.ctype, ast.Symbol("*ret"))],
                         ast.c_for("n", self.cdim,
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3748,6 +3748,9 @@ class Kernel(Cached):
         code = self._ast.gencode() if self._ast else self._code
         return 'Kernel("""%s""", %r)' % (code, self._name)
 
+    def __eq__(self, other):
+        return self.cache_key == other.cache_key
+
 
 class JITModule(Cached):
 
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -448,12 +448,12 @@ class Arg(object):
         return self._is_indirect and self._access is not READ
 
     @property
-    def _is_readonly(self):
+    def _is_read(self):
         return self._access == READ
 
     @property
     def _is_written(self):
-        return not self._is_readonly
+        return not self._is_read
 
     @property
     def _is_indirect_reduction(self):
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -447,6 +447,14 @@ class Arg(object):
         return self._is_indirect and self._access is not READ
 
     @property
+    def _is_readonly(self):
+        return self._access == READ
+
+    @property
+    def _is_written(self):
+        return not self._is_readonly
+
+    @property
     def _is_indirect_reduction(self):
         return self._is_indirect and self._access is INC
 
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -56,6 +56,7 @@ from sparsity import build_sparsity
 from version import __version__ as version
 
 from coffee.base import Node
+from coffee.utils import visit as ast_visit
 from coffee import base as ast
 
 
@@ -3921,6 +3922,15 @@ class ParLoop(LazyComputation):
                     raise RuntimeError("Iteration over a LocalSet does not make sense for RW args")
 
         self._it_space = self.build_itspace(iterset)
+
+        # Attach semantical information to the kernel's AST
+        if hasattr(self._kernel, '_ast') and self._kernel._ast:
+            ast_info = ast_visit(self._kernel._ast, search=ast.FunDecl)
+            fundecl = ast_info['search'][ast.FunDecl]
+            if len(fundecl) == 1:
+                for arg, f_arg in zip(self._actual_args, fundecl[0].args):
+                    if arg._uses_itspace and arg._is_INC:
+                        f_arg.pragma = ast.WRITE
 
     def _run(self):
         return self.compute()
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -1892,10 +1892,10 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         """Zero the data associated with this :class:`Dat`"""
         if not hasattr(self, '_zero_kernel'):
             k = ast.FunDecl("void", "zero",
-                            [ast.Decl(self.ctype, ast.Symbol("*self"))],
+                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self"))],
                             body=ast.c_for("n", self.cdim,
                                            ast.Assign(ast.Symbol("self", ("n", )),
-                                                      ast.FlatBlock("(%s)0" % self.ctype)),
+                                                      ast.Symbol("(%s)0" % self.ctype)),
                                            pragma=None))
             self._zero_kernel = _make_object('Kernel', k, 'zero')
         par_loop(self._zero_kernel, self.dataset.set, self(WRITE))
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -149,13 +149,19 @@ class ExecutionTrace(object):
             else:
                 comp._scheduled = False
 
-        new_trace = list()
+        to_run, new_trace = list(), list()
         for comp in self._trace:
             if comp._scheduled:
-                comp._run()
+                to_run.append(comp)
             else:
                 new_trace.append(comp)
         self._trace = new_trace
+
+        if configuration['loop_fusion']:
+            from fusion import fuse
+            to_run = fuse('from_trace', to_run, 0)
+        for comp in to_run:
+            comp._run()
 
 
 _trace = ExecutionTrace()
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -2020,10 +2020,10 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
                              self.dataset.dim, other.dataset.dim)
 
     def _op(self, other, op):
-        ops = {operator.add: '+',
-               operator.sub: '-',
-               operator.mul: '*',
-               operator.div: '/'}
+        ops = {operator.add: ast.Sum,
+               operator.sub: ast.Sub,
+               operator.mul: ast.Prod,
+               operator.div: ast.Div}
         ret = _make_object('Dat', self.dataset, None, self.dtype)
         name = "binop_%s" % op.__name__
         if np.isscalar(other):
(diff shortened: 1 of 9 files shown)


kynan:
@FabioLuporini Surely firedrakeproject/firedrake#492 depends on this one not vice versa?

pyop2/base.py
```diff
@@ -436,6 +447,14 @@ def _is_indirect_and_not_read(self):
         return self._is_indirect and self._access is not READ
 
     @property
+    def _is_readonly(self):
+        return self._access == READ
+
+    @property
+    def _is_written(self):
```
kynan: The opposite of read-only is writeable, no?

pyop2/base.py
```diff
@@ -3640,17 +3657,22 @@ def name(self):
         """Kernel name, must match the kernel function name in the code."""
         return self._name
 
-    @property
```
kynan: Why is this becoming a method?

pyop2/base.py
```diff
@@ -3640,17 +3657,22 @@ def name(self):
...
+        if not self._code:
+            self._code = self._ast_to_c(self._ast, self._opts)
         return self._code
 
     def __str__(self):
         return "OP2 Kernel: %s" % self._name
 
     def __repr__(self):
-        return 'Kernel("""%s""", %r)' % (self._code, self._name)
+        code = self._ast.gencode() if self._ast else self._code
```
kynan: Why not use `self.code()`?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
+# This file is part of PyOP2
+#
+# PyOP2 is Copyright (c) 2012, Imperial College London and
```
kynan: Side note: this should probably become 2012-2015

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+        """
+
+        def convert(arg, gtl_map, loop_id):
+            # Retrive local maps
+            maps = as_tuple(arg.map, Map)
+            c_local_maps = [None]*len(maps)
+            for i, map in enumerate(maps):
+                c_local_maps[i] = [None]*len(map)
+                for j, m in enumerate(map):
+                    c_local_maps[i][j] = gtl_map["%s%d_%d" % (m.name, i, j)]
```
kynan: Why convert to string instead of using tuple as key?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+        """Given a sequence of tuples of ``Args``, where each tuple comes from a
+        different loop, create a sequence of ``Args`` where there are no duplicates
+        and access modes are properly set (for example, an ``Arg`` whose ``Dat``
+        appears in two different tuples with access mode ``WRITE`` and ``READ``,
+        respectively, will have access mode ``RW`` in the returned sequence of
+        ``Args``."""
+        filtered_args = OrderedDict()
+        for args in loop_args:
+            for a in args:
+                filtered_args[a.data] = filtered_args.get(a.data, a)
```
kynan: aka `filtered_args.setdefault(a.data, a)`

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+                if a.access != filtered_args[a.data].access:
+                    if READ in [a.access, filtered_args[a.data].access]:
+                        # If a READ and some sort of write (MIN, MAX, RW, WRITE,
+                        # INC), then the access mode becomes RW
```
kynan: I don't understand: where is the write?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+                        filtered_args[a.data]._access = RW
+                    elif WRITE in [a.access, filtered_args[a.data].access]:
+                        # Can't be a READ, so just stick to WRITE regardless of what
+                        # the other access mode is
+                        filtered_args[a.data]._access = WRITE
+                    else:
+                        # Neither READ nor WRITE, so access modes are some
+                        # combinations of RW, INC, MIN, MAX. For simplicity,
+                        # just make it RW.
```
kynan: Is that still semantically correct?

pyop2/base.py
FabioLuporini: well, we distinguish between "is_written" and "is_incremented", so isn't it better and much clearer to have properties for each case? as we also have now, incidentally

pyop2/base.py
FabioLuporini: Since we are changing attribute values, I thought not being a property would be much more clear

pyop2/base.py
FabioLuporini: Because otherwise it triggers AST transformation through COFFEE

pyop2/fusion.py
FabioLuporini: Can change

pyop2/fusion.py
FabioLuporini: I'm not getting this

pyop2/fusion.py
FabioLuporini: yes...can improve this

pyop2/fusion.py
FabioLuporini: Uhm, I don't get what you're saying..."in the list", I'd say, there could well be a write,

pyop2/fusion.py
FabioLuporini: INC, MIN, MAX are a more "restrict" case of RW, aren't they? So I'd say yes

pyop2/base.py
```diff
@@ -3826,6 +3849,15 @@ def __init__(self, kernel, iterset, *args, **kwargs):
 
         self._it_space = self.build_itspace(iterset)
 
+        # Attach semantical information to the kernel's AST
```
dham: semantic

pyop2/fusion.py
dham: Florian is pointing out that any hashable python object can be used as a dictionary key so there is no point in converting (m.name, i, j) to a string. You could just as well write:
gtl_map[m.name, i, j]

pyop2/fusion.py
dham: Ah. I think I understand. Florian missed the significance of line 107. Line 107 implies that a is being accessed via two different descriptors. If one of them is a READ then the other must be some form of write, and the union of READ and anything else is RW.

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+                filtered_args[a.data] = filtered_args.get(a.data, a)
+                if a.access != filtered_args[a.data].access:
+                    if READ in [a.access, filtered_args[a.data].access]:
+                        # If a READ and some sort of write (MIN, MAX, RW, WRITE,
+                        # INC), then the access mode becomes RW
+                        filtered_args[a.data]._access = RW
+                    elif WRITE in [a.access, filtered_args[a.data].access]:
+                        # Can't be a READ, so just stick to WRITE regardless of what
+                        # the other access mode is
+                        filtered_args[a.data]._access = WRITE
```
dham: I *think* this is unsafe. Suppose I have a WRITE loop followed by an INC loop. You *might* be able to fuse them but it would require special handling...

pyop2/fusion.py
dham: Yes but INC is incredibly common and it's an important case...

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+    def __init__(self, kernels, fused_ast=None, loop_chain_index=None):
+        """Initialize a :class:`fusion.Kernel` object.
+
+        :param kernels: an iterator of some :class:`Kernel` objects. The objects
+                        can be of class `fusion.Kernel` or of any superclass.
+        :param fused_ast: the Abstract Syntax Tree of the fused kernel. If not
+                          provided, kernels are simply concatenated.
+        :param loop_chain_index: index (i.e., position) of the kernel in a loop
+                                 chain. This can be used to differentiate a same
+                                 kernel appearing multiple times in a loop chain.
```
dham: ...to identify the same kernel...

The current form has a grammar issue and uses the rather overloaded term "differrentiate"

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+        """
+        # Protect against re-initialization when retrieved from cache
+        if self._initialized:
+            return
+
+        asts = fused_ast
+        if not asts:
+            # If kernels' need be concatenated, discard duplicates
+            kernels = dict(zip([k.cache_key[1:] for k in kernels], kernels)).values()
```
dham: Why is discarding duplicates the Right Thing? Maybe we are fusing the same kernel called twice...

pyop2/base.py
dorugeber: @FabioLuporini: @kynan was referring to the name of the property, not its existence/absence.  I.e. if you have `_is_readonly` then you would expect this to be called `_is_writeable` rather than `_is_written`.

dham: arguably it's _is_readonly which is the funny terminology, since the argument is in fact read or written.

pyop2/sequential.py
```diff
@@ -71,10 +71,6 @@ class JITModule(host.JITModule):
     %(buffer_decl)s;
     %(buffer_gather)s
     %(kernel_name)s(%(kernel_args)s);
-    %(layout_decl)s;
-    %(layout_loop)s
-        %(layout_assign)s;
-    %(layout_loop_close)s
```
dorugeber: Where did these go?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+            self._info[pos]['loop_indices'] = [base_idx, fuse_idx]
+            # In addition to the union of the /base/ and /fuse/'s sets of arguments,
+            # a bitmap, with i-th bit indicating whether the i-th iteration in "fuse"
+            # has been executed, will have to be passed in
+            self._info[pos]['extra_args'] = [((fused_map.toset, None, np.int32),
+                                              (RW, fused_map))]
+            kernel.pop(pos+1)
+            pos = max(base_idx, fuse_idx)
+            self._info.pop(pos)
+            kernel.pop(pos)
```
kynan: What is being removed here?

pyop2/base.py
kynan: Indeed. `_is_read` and `_is_written` would be preferable I think.

pyop2/base.py
kynan: You mean because it mutates the state of the object?

pyop2/fusion.py
kynan: Maybe don't, better to do it as a single change over the entire codebase.

pyop2/fusion.py
kynan: OK yes, now it's clear.

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+        self._kernel = kernel
+
+    def __call__(self, loop_chain, only_hard=False):
+        # First apply soft fusion, then hard fusion
+        if not only_hard:
+            loop_chain = self._schedule(loop_chain)
```
kynan: What is happening here and why is this equivalent to soft fusion?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+            # this inspection
+            if self.mode != mode:
+                raise RuntimeError("Cached Inspector's mode doesn't match")
+            return self._schedule
+        elif not hasattr(self, '_loop_chain'):
+            # The inspection should be executed /now/. We weren't in the cache,
+            # so we /must/ have a loop chain
+            raise RuntimeError("Inspector must have a loop chain associated with it")
+        # Finally, we check the legality of `mode`
+        if mode not in Inspector._modes:
```
kynan: Shouldn't you check this first before you even compute the heuristic?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+        del self._loop_chain
+        del self._tile_size
+        return self._schedule
+
+    def _heuristic_skip_inspection(self, mode):
+        """Decide heuristically whether to run an inspection or not."""
+        # At the moment, a simple heuristic is used. If tiling is not requested,
+        # then inspection and fusion are always performed. If tiling is on the other
+        # hand requested, then fusion is performed only if inspection is requested
+        # more than once. This is to amortize the cost of inspection due to tiling.
```
kynan: Why would inspection be requested multiple times?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+                     dat_1_2(READ, ...),
+                     ...)
+
+            par_loop(kernel_2, it_space_2,
+                     dat_2_1(INC, ...),
+                     dat_2_2(READ, ...),
+                     ...)
+
+        where ``dat_1_1 == dat_2_1`` and, possibly (but not necessarily),
+        ``it_space_1 != it_space_2``, can be hardly fused. Note, in fact, that
```
kynan: I don't think you want to use hardly here.

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+        return loop_chain
+
+    # Search for _Assembly objects since they introduce a synchronization point;
+    # that is, loops cannot be fused across an _Assembly object. In that case, try
+    # to fuse only the segment of loop chain right before the synchronization point
+    remainder = []
+    synch_points = [l for l in loop_chain if isinstance(l, Mat._Assembly)]
+    if synch_points:
+        if len(synch_points) > 1:
+            warning("Fusing loops and found more than one synchronization point")
```
kynan: Is a warning enough here? afaiu you only handle the 1st synchronisation point. Is this still semantically correct with > 1?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+        synch_point = loop_chain.index(synch_points[0])
+        remainder, loop_chain = loop_chain[synch_point:], loop_chain[:synch_point]
```
kynan: Swapping the tuple elements would make this slightly easier to understand.

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+    if any([l.is_layered for l in loop_chain]):
+        return loop_chain + remainder
+
+    # Check if tiling needs be applied
+    mode = 'hard'
+    if tile_size > 0:
+        mode = 'tile'
+        # Loop tiling requires the SLOPE library to be available on the system.
+        try:
+            os.environ['SLOPE_DIR']
```
kynan: Are you planning to improve the build/installation process of SLOPE so this wouldn't have to be set manually?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+        [fused_loops_0, fused_loops_1, ..., fused_loops_m-1, peel_loops]
+
+    which is eventually inserted in the global trace of :class:`ParLoop` objects.
+
+    That is, sub-sequences of :class:`ParLoop` objects are potentially replaced by
+    new :class:`ParLoop` objects representing the fusion or the tiling of the
+    original trace slice.
+
+    :param name: identifier of the loop chain
+    :param time_unroll: in a time stepping loop, the length of the loop chain
```
kynan: Better call this `num_unroll`: this is an int, not a time value.

pyop2/base.py
FabioLuporini: I agree, will change

pyop2/base.py
FabioLuporini: Yes

pyop2/fusion.py
FabioLuporini: Ah right, yes. I put it as a string because later one uses that string, so sooner or later the conversion will happen anyway. But I can postpone the conversion if you feel it'd be much cleaner

kynan: No, leave it as is.

kynan:
I skipped over the actual fusion code (I don't claim to understand what's going on there). The rest looks mostly sensible afaict.

pyop2/fusion.py
FabioLuporini: Yes, but look at this line 107; a new access mode is set *iff* the access modes of two arguments with same payload are different. So, if they are both ``INC``, the access mode actually remains ``INC``.  

The comment "For simplicity..." is just to say "instead of having the product of all of the possible access modes (RW_INC, RW_MIN, etc.), which would probably be useless, let's just stick to the general one, RW"

pyop2/fusion.py
FabioLuporini: This function should be used (it is) to set the access modes in the ``fused parloop``, not for data dependency analysis, which is performed in the ``Inspector``, and is specific to ``soft_fusion``, ``hard_fusion``, and ``tiling``.

pyop2/fusion.py
FabioLuporini: We are just discarding duplicate code. Of course the same kernel can appear multiple times in the fused parloop: this is totally supported, but this "discarding thing" is just to avoid that the same piece of C code uselessly appears multiple times

pyop2/sequential.py
FabioLuporini: so this branch is rebased on ``remove_host_buffer`` and this change reflects the fact that now coffee does padding without the help of PyOP2

pyop2/fusion.py
FabioLuporini: We are replacing original ParLoops with a fused ParLoop, so I'm removing anything regarding the original ParLoops

pyop2/fusion.py
FabioLuporini: The schedules are applied incrementally. Given a loop chain, you first apply soft fusion, then hard fusion (then tiling is specifically requested by the user, but this is another story). 

``only_hard`` is to invoke the scheduler without performing soft_fusion on the provided loop chain because you already know that it's been performed already. This is the case when you have just created a hard fusion ParLoop, and you want to update the (already soft fused) loop chain

pyop2/fusion.py
FabioLuporini: Yes, good spot, I'll move it

pyop2/fusion.py
FabioLuporini: Because for example you are in a time loop. So you would inspect the same sequence of loops multiple times - but since we cache it, inspection happens only once

pyop2/fusion.py
FabioLuporini: Are you suggesting the terms "hard" and "soft" don't reflect what's happening? Could you elaborate?

pyop2/fusion.py
FabioLuporini: Yes, it should be correct, since you're not throwing away anything. It's just that we don't try to fuse after the second synch point. I think this is plausible, because in practice it should never occur

pyop2/fusion.py
FabioLuporini: I can do this, but what are you suggesting specifically?

pyop2/fusion.py
FabioLuporini: I agree, will change

pyop2/base.py
miklos1: The pattern you are using here is implemented as `utils.cached_property` in Firedrake. I think it would be clearer to just use that.

pyop2/base.py
miklos1: No.
This returns a different representation based on whether the property `code` was already calculated (and cached) or not. Sounds evil. I think it is better to choose one or the other as representation and to use solely that representation.

pyop2/fusion.py
FabioLuporini: I've tried fixing a few but it looks like I'm missing a rule here - could you come up with a couple of examples?

wence-: both are fine.  It depends if you believe the kernel belongs to the parloop (the former) or not (the latter).

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+                    c_local_maps[i][j] = gtl_map["%s%d_%d" % (m.name, i, j)]
+            # Instantiate and initialize new, specialized Arg
+            _arg = Arg(arg.data, arg.map, arg.idx, arg.access, arg._flatten)
+            _arg._loop_position = loop_id
+            _arg._position = arg._position
+            _arg._indirect_position = arg._indirect_position
+            _arg._c_local_maps = c_local_maps
+            return _arg
+
+        if isinstance(args, (list, tuple)):
```
miklos1: Alternatively, you may simply try to iterate over `args`, and fall back on handling the single element case, if 'not iterable' exception is caught.

pyop2/fusion.py
miklos1: Why would "the same piece of C code uselessly appear multiple times"?

wence-:
```python
par_loop(kernel_A, ...)
par_loop(kernel_B, ...)
par_loop(kernel_A, ...)
```

The definition of kernel_A should only appear once, but if you do the naivest thing it appears twice?

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+# Parallel loop API
+
+class JITModule(host.JITModule):
+
+    _cppargs = []
+    _libraries = []
+    _extension = 'cpp'
+
+    _wrapper = """
+extern "C" void %(wrapper_name)s(%(executor_arg)s,
```
miklos1: Is this template really new? Don't we have it somewhere in the software stack already? Just asking to avoid potential duplication...

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+                      %(const_args)s) {
+  %(user_code)s
+  %(wrapper_decs)s;
+  %(const_inits)s;
+
+  %(executor_code)s;
+}
+"""
+    _kernel_wrapper = """
+%(interm_globals_decl)s;
```
miklos1: Is this template really new? Don't we have it somewhere in the software stack already? Just asking to avoid potential duplication...

pyop2/fusion.py
```diff
@@ -0,0 +1,1264 @@
...
+    @classmethod
+    def _cache_key(cls, kernel, it_space, *args, **kwargs):
+        key = (hash(kwargs['executor']),)
+        all_args = kwargs['all_args']
+        for kernel_i, it_space_i, args_i in zip(kernel, it_space, all_args):
+            key += super(JITModule, cls)._cache_key(kernel_i, it_space_i, *args_i)
+        return key
+
+    def __init__(self, kernel, it_space, *args, **kwargs):
+        if self._initialized:
```
miklos1: Is it really necessary? Is there a case when `__init__` might be called multiple times on the same object?

wence-: If you pull an object out of the cache in `__new__` the `__init__` method is always called, this way you don't reinit objects coming out the cache.

miklos1: I see, thanks.

pyop2/base.py
FabioLuporini: I can stick to ast.gencode if you prefer, otherwise I'm open to suggestions

pyop2/fusion.py
FabioLuporini: Correct.

pyop2/fusion.py
FabioLuporini: It's new

pyop2/fusion.py
FabioLuporini: It's new

pyop2/fusion.py
kynan: Indeed, I think that'd be better.

miklos1: Bonus question: is this code ever called with a list or a tuple? The only use I could find already loops over a set of elements outside the call.

dorugeber: This sounds like you're saying "I have no idea what args is.  It could be a string, a tuple, a list, a doughnut, a giraffe, ..."

pyop2/fusion.py
```diff
@@ -0,0 +1,1263 @@
...
+        for i, loop in enumerate(self._loop_chain[1:]):
+            base_loop = fusing[-1]
+            if base_loop.it_space != loop.it_space or \
+                    (base_loop.is_indirect and loop.is_indirect):
+                # Fusion not legal
+                fused.append((fuse(self, fusing, len(fused)), i+1))
+                fusing = [loop]
+            elif (base_loop.is_direct and loop.is_direct) or \
+                    (base_loop.is_direct and loop.is_indirect) or \
+                    (base_loop.is_indirect and loop.is_direct):
```
dorugeber: `elif not (base_loop.is_indirect and loop.is_indirect)`?

dorugeber: `elif base_loop.is_direct or loop.is_direct`?

pyop2/fusion.py
```diff
@@ -0,0 +1,1263 @@
...
+        fundecl.args = new_fused_kernel_args
+        return binding
+
+    def _soft_fuse(self):
+        """Fuse consecutive loops over the same iteration set by concatenating
+        kernel bodies and creating new :class:`ParLoop` objects representing
+        the fused sequence.
+
+        The conditions under which two loops over the same iteration set can
+        be hard fused are:
```
wence-: So here you talk about hard fusion, and yet the function name is soft_fuse.  I'm confused.

pyop2/fusion.py
FabioLuporini: I just find cleaner listing the various combinations, it helps me reasoning about the interaction between the two loops

pyop2/fusion.py
FabioLuporini: Agree, changed

pyop2/fusion.py
FabioLuporini: uops, I meant soft, changed

FabioLuporini:
Apart from very minor things, I think I've addressed most of the comments. There were sensible comments from @dham about the semantics and correctness of this, rather than mere implementation details, which I think I answered, but I'd like to hear back from him about my answers. Thanks!

pyop2/fusion.py
dham: OK. Yes. that's fine.

pyop2/fusion.py
dham: OK

dham:
Hi Fabio. I'm happy with the responses you gave.

FabioLuporini:
Latest commit fixes the issue raised by David yesterday about importing SLOPE only if tiling is requested and when needed

PR #434, opened 2015-04-17
Remove Python 2.6 support

wence-:
Some of PyOP2's dependencies (in particular COFFEE) are not 2.6
compatible.  2.7 is widespread enough.


miklos1:
I agree.

kynan:
Agreed.

PR #435, opened 2015-04-20
Rework overhead reduction

wence-:
Here's a reworked lot of reduction in the time we take to do nothing.  It works with lazy (loop fusion, yay!) and gives reasonable speedups for small problems.  I would like confirmation that this works properly for the loop fusion cases: there are no tests in pyop2 of this functionaliy.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3922,30 +3922,41 @@ class ParLoop(LazyComputation):
         """
         return ()
 
+    @property
+    @collective
+    def _jitmodule(self):
+        """Return the :class:`JITModule` that encapsulates the compiled par_loop code.
+
+        Return None if the child class should deal with this in another way."""
+        return None
+
     @collective
     def compute(self):
         """Executes the kernel over all members of the iteration space."""
         self.halo_exchange_begin()
         iterset = self.iterset
         arglist = self.prepare_arglist(iterset, *self.args)
-        self._compute(iterset.core_part, *arglist)
+        fun = self._jitmodule
+        self._compute(iterset.core_part, fun, *arglist)
         self.halo_exchange_end()
-        self._compute(iterset.owned_part, *arglist)
+        self._compute(iterset.owned_part, fun, *arglist)
         self.reduction_begin()
         if self._only_local:
             self.reverse_halo_exchange_begin()
             self.reverse_halo_exchange_end()
         if not self._only_local and self.needs_exec_halo:
-            self._compute(iterset.exec_part, *arglist)
+            self._compute(iterset.exec_part, fun, *arglist)
         self.reduction_end()
         self.maybe_set_halo_update_needed()
         self.maybe_set_dat_dirty()
 
     @collective
-    def _compute(self, part, *arglist):
+    def _compute(self, part, fun, *arglist):
         """Executes the kernel over all members of a MPI-part of the iteration space.
 
         :arg part: The :class:`SetPartition` to compute over
+        :arg fun: The :class:`JITModule` encapsulating the compiled
+             code (may be ignored by the backend).
         :arg arglist: The arguments to pass to the compiled code (may
              be ignored by the backend, depending on the exact implementation)"""
         raise RuntimeError("Must select a backend")
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3914,6 +3914,14 @@ class ParLoop(LazyComputation):
     def _run(self):
         return self.compute()
 
+    def prepare_arglist(self, iterset, *args):
+        """Prepare the argument list for calling generated code.
+
+        :arg iterset: The :class:`Set` iterated over.
+        :arg args: A list of :class:`Args`, the argument to the :fn:`par_loop`.
+        """
+        return ()
+
     @collective
     def compute(self):
         """Executes the kernel over all members of the iteration space."""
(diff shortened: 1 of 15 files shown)


pyop2/base.py
```diff
@@ -256,17 +254,15 @@ def __init__(self, data=None, map=None, idx=None, access=None, flatten=False):
            defined on.
 
         A :class:`MapValueError` is raised if these conditions are not met."""
-        self._dat = data
+        self.data = data
```
dorugeber: I assume there's a good reason for renaming (which is orthogonal to performance)?

pyop2/base.py
```diff
@@ -335,131 +331,109 @@ def split(self):
...
-        """The position of the first unique occurence of this
-    indirect :class:`Arg` in the :class:`ParLoop` argument list."""
-        return self._indirect_position
-
-    @indirect_position.setter
-    def indirect_position(self, val):
-        """Set the position of the first unique occurence of this
-    indirect :class:`Arg` in the :class:`ParLoop` argument list."""
-        self._indirect_position = val
+        return "arg%d" % self.position
```
dorugeber: I assume there's a good reason for these to die, though I have no idea why!

pyop2/base.py
```diff
@@ -2864,46 +2840,39 @@ def __len__(self):
...
         will always be ALL. For a :class:`DecoratedMap` it will specify over which mesh
         region the iteration will take place."""
         return frozenset([ALL])
 
-    @property
-    def implicit_bcs(self):
-        """Return any implicit (extruded "top" or "bottom") bcs to
-        apply to this :class:`Map`.  Normally empty except in the case of
-        some :class:`DecoratedMap`\s."""
-        return frozenset([])
```
dorugeber: Ditto not sure why this dies, but okay.

pyop2/openmp.py
```diff
@@ -206,96 +241,67 @@ def generate_code(self):
...
+                arglist.append(1)
+            elif region is ON_TOP:
+                arglist.append(iterset.layers - 2)
+                arglist.append(iterset.layers - 1)
+            elif region is ON_INTERIOR_FACETS:
+                arglist.append(0)
+                arglist.append(iterset.layers - 2)
+            else:
+                arglist.append(0)
+                arglist.append(iterset.layers - 1)
```
dorugeber: Ugh.  I assume these are 'iterate from a to b'?  Seems a bit too magic for my liking... append two integers on to `arglist` and these just happen to be the loop bounds.

dorugeber: ...I appreciate it was no better before you touched this part of the code!

pyop2/sequential.py
```diff
@@ -79,75 +75,89 @@ class JITModule(host.JITModule):
...
+            elif region is ON_TOP:
+                arglist.append(iterset.layers - 2)
+                arglist.append(iterset.layers - 1)
+            elif region is ON_INTERIOR_FACETS:
+                arglist.append(0)
+                arglist.append(iterset.layers - 2)
+            else:
+                arglist.append(0)
+                arglist.append(iterset.layers - 1)
+        return arglist
```
dorugeber: Ditto.

test/unit/test_versioning.py
```diff
@@ -111,7 +111,7 @@ def mat(cls, iter2ind1):
         return op2.Mat(sparsity, 'float64', "mat")
 
     def test_initial_version(self, backend, skip_opencl, mat, g, x):
-        assert mat._version == 1
+        assert mat._version == 0
```
dorugeber: I assume everything else starts with version 0 and this is now consistent?

dorugeber:
Not obviously insane.

pyop2/base.py
wence-: There was always a public `arg.data` property.  This just gives only the "public" property.

pyop2/base.py
wence-: So property access is slower than slot access, and every time we build a par_loop we set these values.  So now we just set the public slots directly, rather than going through the property slots.

pyop2/base.py
wence-: I shall make it a cached_property.

pyop2/openmp.py
wence-: Well, you need to get the layer bounds into the wrapper code somehow.  You've either done it by hard-coding them, or you pass them in here.

test/unit/test_versioning.py
wence-: I /think/ so.

pyop2/openmp.py
dorugeber: But there's no explicit indication that these are the layer bounds!  Had the code been
```
arglist.append(("lower_bound", 0))
arglist.append(("upper_bound", 1))
```
then it would have been clear.  But it looks like the bounds are arbitrarily inserted into the middle of a list, and I guess this list is unpicked later on?

wence-: So this is exactly just the argument list that's passed to the compiled code.  I don't want to build and then unpick it.  I shall add some comments.

pyop2/base.py
```diff
@@ -2786,6 +2747,19 @@ def __iter__(self):
 """
 
 
+class _MapArg(object):
```
kynan: Since these are immutable, could you have used a [named tuple](https://docs.python.org/2/library/collections.html#collections.namedtuple) instead?

wence-: FWIW, which isn't much, named tuples are slightly slower.  ballpark 300ns per object instantiation as opposed to 250ns.

pyop2/fusion.py
```diff
@@ -267,17 +267,31 @@ def __init__(self, kernel, it_space, *args, **kwargs):
...
-    def compile(self, argtypes=None, restype=None):
-        if hasattr(self, '_fun'):
-            # It should not be possible to pull a jit module out of
-            # the cache /with/ arguments
-            if hasattr(self, '_args'):
-                raise RuntimeError("JITModule is holding onto args, memory leak!")
-            self._fun.argtypes = argtypes
-            self._fun.restype = restype
-            return self._fun
+    def set_argtypes(self, iterset, *args):
```
kynan: This method doesn't really set anything, could be called `build_argtypes` rather.

kynan:
Out of interest: did you measure the difference between `property` and `cached_property`?

PR #436, opened 2015-04-21
Use AST representation in host.py (Lower layers loop branch).

doru1004:
In the extruded and non-extruded cases, transform the following template strings to be represented using COFFEE ASTs for both sequential and OpenMP backends:

```
    ...
    %(extr_loop)s
    %(map_bcs_m)s;
    %(buffer_decl)s;
    %(buffer_gather)s
    %(kernel_name)s(%(kernel_args)s);
    %(itset_loop_body)s
    %(map_bcs_p)s;
    %(apply_offset)s;
    %(extr_loop_close)s
   ...
```

When the kernel AST does not contain any AST nodes of type FlatBlock, inline the kernel code into the wrapper and rename variables to avoid clashes. (TODO: relax the FlatBlock requirement for static declarations).

Fixes: In declarations using ASTs, associate the pointer with the type instead of the actual variable symbol.

This branch depends on similarly named branches in:
Firedrake: https://github.com/firedrakeproject/firedrake/pull/501
COFFEE: https://github.com/coneoproject/COFFEE/pull/41

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -1911,7 +1911,7 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
                             [ast.Decl("%s*" % self.ctype, ast.Symbol("self"))],
                             body=ast.c_for("n", self.cdim,
                                            ast.Assign(ast.Symbol("self", ("n", )),
-                                                      ast.Symbol("(%s)0" % self.ctype)),
+                                                      ast.Cast(self.ctype, ast.Symbol("0"))),
                                            pragma=None))
             self._zero_kernel = _make_object('Kernel', k, 'zero')
         par_loop(self._zero_kernel, self.dataset.set, self(WRITE))
@@ -2034,7 +2034,7 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
                                       qualifiers=["const"]),
                              ast.Decl(other.ctype + "*", ast.Symbol("other"),
                                       qualifiers=["const"]),
-                             ast.Decl(self.ctype, ast.Symbol("*ret"))],
+                             ast.Decl(self.ctype + "*", ast.Symbol("ret"))],
                             ast.c_for("n", self.cdim,
                                       ast.Assign(ast.Symbol("ret", ("n", )),
                                                  ops[op](ast.Symbol("self", ("n", )),
@@ -3773,17 +3773,19 @@ class JITModule(Cached):
                 else:
                     idx = arg.idx
                 map_arity = arg.map.arity if arg.map else None
-                key += (arg.data.dim, arg.data.dtype, map_arity, idx, arg.access)
+                map_offset = str(arg.map.offset) if arg.map else ""
+                key += (arg.data.dim, arg.data.dtype, map_arity, idx, arg.access, map_offset)
             elif arg._is_mat:
                 idxs = (arg.idx[0].__class__, arg.idx[0].index,
                         arg.idx[1].index)
                 map_arities = (arg.map[0].arity, arg.map[1].arity)
+                map_offsets = (str(arg.map[0].offset), str(arg.map[1].offset))
                 # Implicit boundary conditions (extruded "top" or
                 # "bottom") affect generated code, and therefore need
                 # to be part of cache key
                 map_bcs = (arg.map[0].implicit_bcs, arg.map[1].implicit_bcs)
                 key += (arg.data.dims, arg.data.dtype, idxs,
-                        map_arities, map_bcs, arg.access)
+                        map_arities, map_bcs, arg.access, map_offsets)
 
         iterate = kwargs.get("iterate", None)
         if iterate is not None:
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -1931,9 +1931,9 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         """Create the :class:`ParLoop` implementing copy."""
         if not hasattr(self, '_copy_kernel'):
             k = ast.FunDecl("void", "copy",
-                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self"),
+                            [ast.Decl(self.ctype + "*", ast.Symbol("self"),
                                       qualifiers=["const"]),
-                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"))],
+                             ast.Decl(other.ctype + "*", ast.Symbol("other"))],
                             body=ast.c_for("n", self.cdim,
                                            ast.Assign(ast.Symbol("other", ("n", )),
                                                       ast.Symbol("self", ("n", ))),
@@ -2030,9 +2030,9 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         if np.isscalar(other):
             other = _make_object('Global', 1, data=other)
             k = ast.FunDecl("void", name,
-                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self"),
+                            [ast.Decl(self.ctype + "*", ast.Symbol("self"),
                                       qualifiers=["const"]),
-                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
+                             ast.Decl(other.ctype + "*", ast.Symbol("other"),
                                       qualifiers=["const"]),
                              ast.Decl(self.ctype, ast.Symbol("*ret"))],
                             ast.c_for("n", self.cdim,
@@ -2045,11 +2045,11 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         else:
             self._check_shape(other)
             k = ast.FunDecl("void", name,
-                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self"),
+                            [ast.Decl(self.ctype + "*", ast.Symbol("self"),
                                       qualifiers=["const"]),
-                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
+                             ast.Decl(other.ctype + "*", ast.Symbol("other"),
                                       qualifiers=["const"]),
-                             ast.Decl("%s*" % self.ctype, ast.Symbol("ret"))],
+                             ast.Decl(self.ctype + "*", ast.Symbol("ret"))],
                             ast.c_for("n", self.cdim,
                                       ast.Assign(ast.Symbol("ret", ("n", )),
                                                  ops[op](ast.Symbol("self", ("n", )),
@@ -2070,8 +2070,8 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         if np.isscalar(other):
             other = _make_object('Global', 1, data=other)
             k = ast.FunDecl("void", name,
-                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self")),
-                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
+                            [ast.Decl(self.ctype + "*", ast.Symbol("self")),
+                             ast.Decl(other.ctype + "*", ast.Symbol("other"),
                                       qualifiers=["const"])],
                             ast.c_for("n", self.cdim,
                                       ops[op](ast.Symbol("self", ("n", )),
@@ -2082,8 +2082,8 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
             self._check_shape(other)
             quals = ["const"] if self is not other else []
             k = ast.FunDecl("void", name,
-                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self")),
-                             ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
+                            [ast.Decl(self.ctype + "*", ast.Symbol("self")),
+                             ast.Decl(other.ctype + "*", ast.Symbol("other"),
                                       qualifiers=quals)],
                             ast.c_for("n", self.cdim,
                                       ops[op](ast.Symbol("self", ("n", )),
@@ -2097,7 +2097,7 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         ops = {operator.sub: ast.Neg}
         name = "uop_%s" % op.__name__
         k = ast.FunDecl("void", name,
-                        [ast.Decl("%s*" % self.ctype, ast.Symbol("self"))],
+                        [ast.Decl(self.ctype + "*", ast.Symbol("self"))],
                         ast.c_for("n", self.cdim,
                                   ast.Assign(ast.Symbol("self", ("n", )),
                                              ops[op](ast.Symbol("self", ("n", )))),
@@ -2117,11 +2117,11 @@ class Dat(SetAssociated, _EmptyDataMixin, CopyOnWrite):
         ret = _make_object('Global', 1, data=0, dtype=self.dtype)
 
         k = ast.FunDecl("void", "inner",
-                        [ast.Decl("%s*" % self.ctype, ast.Symbol("self"),
+                        [ast.Decl(self.ctype + "*", ast.Symbol("self"),
                                   qualifiers=["const"]),
-                         ast.Decl("%s*" % other.ctype, ast.Symbol("other"),
+                         ast.Decl(other.ctype + "*", ast.Symbol("other"),
                                   qualifiers=["const"]),
-                         ast.Decl(self.ctype, ast.Symbol("*ret"))],
+                         ast.Decl(self.ctype + "*", ast.Symbol("ret"))],
                         ast.c_for("n", self.cdim,
                                   ast.Incr(ast.Symbol("ret", (0, )),
                                            ast.Prod(ast.Symbol("self", ("n", )),
--- a/pyop2/host.py
+++ b/pyop2/host.py
@@ -929,6 +929,44 @@ class JITModule(base.JITModule):
                 return "{"
             return "for (int j_0 = start_layer; j_0 < end_layer; ++j_0){"
 
+        def get_kernel_decl_node(kernel_ast):
+            # Find the reference to node continaing the kernel header
+            if isinstance(kernel_ast, ast.FunDecl):
+                return kernel_ast
+            if kernel_ast and not isinstance(kernel_ast, str):
+                for child in kernel_ast.children:
+                    c = get_kernel_decl_node(child)
+                    if isinstance(c, ast.FunDecl):
+                        return c
+            return None
+
+        def rename_var(kernel_ast, old_name, new_name):
+            # Because the AST representation is not consistent
+            # this function might fail. FFC pollutes names of variables with
+            # *, & and [] symbols which SHOULD NOT BE PRESENT in the name of a variable.
+            if isinstance(kernel_ast, ast.Symbol):
+                var_name = str(kernel_ast.symbol)
+                if "*" in var_name or "[" in var_name or "]" in var_name or "&" in var_name:
+                    raise RuntimeError("The variable name has been polluted with * & [ or ] characters.")
+                if var_name == old_name:
+                    print "2", kernel_ast.symbol, new_name
+                    kernel_ast.symbol = str(kernel_ast.symbol).replace(old_name, new_name)
+            elif not isinstance(kernel_ast, str):
+                print "3"
+                for child in kernel_ast.children:
+                    rename_var(child, old_name, new_name)
+            print "4"
+
+        def contains_flat_block(tree):
+            # Check if the AST contains any FlatBlock nodes
+            if isinstance(tree, ast.FlatBlock):
+                return True
+            elif not isinstance(tree, str):
+                for child in tree.children:
+                    if contains_flat_block(child):
+                        return True
+            return False
+
         _ssinds_arg = ""
         _index_expr = "n"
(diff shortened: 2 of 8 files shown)


pyop2/host.py
```diff
@@ -187,6 +222,47 @@ def c_kernel_arg(self, count, i=0, j=0, shape=(0,), is_top=False, layers=1):
...
+        if self._uses_itspace:
+            if self._is_mat:
+                if self.data[i, j]._is_vector_field:
+                    return ast.Symbol(self.c_kernel_arg_name(i, j))
+                elif self.data[i, j]._is_scalar_field:
+                    flat_block = "(%(t)s (*)[%(dim)d])&%(name)s" % \
+                        {'t': self.ctype,
+                         'dim': shape[0],
+                         'name': self.c_kernel_arg_name(i, j)}
+                    return ast.FlatBlock(flat_block)
```
FabioLuporini: We want to avoid FlatBlocks always, when possible - and this is one of those cases

pyop2/host.py
```diff
@@ -234,17 +310,17 @@ def c_vec_init(self, is_top, layers, is_facet=False):
...
-    def c_addto(self, i, j, buf_name, tmp_name, tmp_decl,
-                extruded=None, is_facet=False, applied_blas=False):
+    def ast_addto(self, i, j, buf_name, tmp_name, tmp_decl,
+                  extruded=None, is_facet=False, applied_blas=False):
         maps = as_tuple(self.map, Map)
         nrows = maps[0].split[i].arity
         ncols = maps[1].split[j].arity
-        rows_str = "%s + i * %s" % (self.c_map_name(0, i), nrows)
-        cols_str = "%s + i * %s" % (self.c_map_name(1, j), ncols)
+        rows_ast = ast.Sum(ast.Symbol(self.c_map_name(0, i)), ast.Prod(ast.Symbol("i"), ast.Symbol(nrows)))
```
FabioLuporini: Literals are mapped to symbols implicitly, and this works well if you don't have ranks or offsets.
So for example ast.Prod(ast.Symbol("i"), ast.Symbol(nrows)) == ast.Prod('i', nrows)

pyop2/host.py
```diff
@@ -860,77 +1039,107 @@ def extrusion_loop():
...
+            change_kernel_loop_index_ids(kernel_decl)
+            for i, arg in enumerate(kernel_decl.args):
+                if not isinstance(_kernel_args[i], ast.Symbol):
+                    arg.init = _kernel_args[i]
+                    extr_loop_body += [arg]
+                else:
+                    rename_var(kernel_decl, str(arg.sym.symbol), str(_kernel_args[i].symbol))
+            extr_loop_body += kernel_decl.children
+        else:
+            extr_loop_body += [ast.FlatBlock("%(name)s(%(args)s);\n" % {'name': self._kernel.name, 'args': _kernel_args_str})]
```
FabioLuporini: Let's avoid FlatBlocks

test/unit/test_matrices.py
```diff
@@ -787,7 +787,7 @@ def mat(self, msparsity, mmap, mdat):
                            Incr(Symbol("v", ("i", "j")), FlatBlock("d[i][0] * d[j][0]"))))
```
FabioLuporini: This FlatBlock here should also go

pyop2/base.py
```diff
@@ -1931,9 +1931,9 @@ def _copy_parloop(self, other, subset=None):
         """Create the :class:`ParLoop` implementing copy."""
         if not hasattr(self, '_copy_kernel'):
             k = ast.FunDecl("void", "copy",
-                            [ast.Decl("%s*" % self.ctype, ast.Symbol("self"),
+                            [ast.Decl(self.ctype + "*", ast.Symbol("self"),
```
kynan: Any particular reason for this (and the following) spurious changes?

pyop2/host.py
```diff
@@ -133,15 +135,48 @@ def c_ind_data(self, idx, i, j=0, is_top=False, layers=1, offset=None):
...
+        # return "%(name)s + (%(map_name)s[i * %(arity)s + %(idx)s]%(top)s%(off_mul)s%(off_add)s)* %(dim)s%(off)s" % \
+        #     {'name': self.c_arg_name(i),
+        #      'map_name': self.c_map_name(i, 0),
+        #      'arity': self.map.split[i].arity,
+        #      'idx': idx,
+        #      'top': ' + start_layer' if is_top else '',
+        #      'dim': self.data[i].cdim,
+        #      'off': ' + %d' % j if j else '',
+        #      'off_mul': ' * %d' % offset if is_top and offset is not None else '',
+        #      'off_add': ' + %d' % offset if not is_top and offset is not None else ''}
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -133,15 +135,48 @@ def c_ind_data(self, idx, i, j=0, is_top=False, layers=1, offset=None):
...
+            tail = ast.Sum(tail, ast.Symbol(j))
+        return ast.Sum(ast.Symbol(self.c_arg_name(i)), tail)
+        # return "%(name)s + (xtr_%(map_name)s[%(idx)s]%(top)s%(offset)s)*%(dim)s%(off)s" % \
+        #     {'name': self.c_arg_name(i),
+        #      'map_name': self.c_map_name(i, 0),
+        #      'idx': idx,
+        #      'top': ' + start_layer' if is_top else '',
+        #      'dim': 1 if self._flatten else str(self.data[i].cdim),
+        #      'off': ' + %d' % j if j else '',
+        #      'offset': ' * _'+self.c_offset_name(i, 0)+'['+idx+']' if is_top else ''}
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -187,6 +222,47 @@ def c_kernel_arg(self, count, i=0, j=0, shape=(0,), is_top=False, layers=1):
...
+                    index = ast.Sum(index, ast.Mod(ast.Symbol("i_0"), ast.Symbol(self.map.arity)))
+                    tail = ast.Symbol(self.c_map_name(i, 0), rank=(index,))
+                    tail = ast.Prod(tail, ast.Symbol(self.data[i].cdim))
+                    tail = ast.Sum(tail, ast.Div(ast.Symbol("i_0"), ast.Symbol(self.map.arity)))
+                    return ast.Sum(ast.Symbol(self.c_arg_name()), tail)
+                    # return "%(name)s + %(map_name)s[i * %(arity)s + i_0 %% %(arity)d] * %(dim)s + (i_0 / %(arity)d)" % \
+                    #     {'name': self.c_arg_name(),
+                    #      'map_name': self.c_map_name(0, i),
+                    #      'arity': self.map.arity,
+                    #      'dim': self.data[i].cdim}
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -326,31 +399,43 @@ def c_zero_tmp(self, i, j):
...
+                    # rhs = ast.Symbol(self.c_offset_name(i, 0), rank=(idx,))
+                    # rhs = ast.Prod(rhs, ast.Symbol(d.dataset.cdim))
+                    val_ast.append(ast.Incr(lhs, rhs))
+                    # val.append("%(name)s[%(j)d] += %(off)d * %(dim)s;" %
+                    #            {'name': self.c_vec_name(),
+                    #             'i': idx,
+                    #             'j': vec_idx,
+                    #             'offset': self.c_offset_name(i, 0),
+                    #             'dim': d.dataset.cdim,
+                    #             'off': m.offset[idx]})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -326,31 +399,43 @@ def c_zero_tmp(self, i, j):
...
-                    val.append("%(name)s[%(j)d] += %(offset)s[%(i)d] * %(dim)s;" %
-                               {'name': self.c_vec_name(),
-                                'i': idx,
-                                'j': vec_idx,
-                                'offset': self.c_offset_name(i, 0),
-                                'dim': d.dataset.cdim})
+                    lhs = ast.Symbol(self.c_vec_name(), rank=(vec_idx,))
+                    rhs = ast.Symbol(m.offset[idx] * d.dataset.cdim)
+                    # rhs = ast.Symbol(self.c_offset_name(i, 0), rank=(idx,))
+                    # rhs = ast.Prod(rhs, ast.Symbol(d.dataset.cdim))
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -326,31 +399,43 @@ def c_zero_tmp(self, i, j):
...
-                        val.append("%(name)s[%(j)d] += %(offset)s[%(i)d] * %(dim)s;" %
-                                   {'name': self.c_vec_name(),
-                                    'i': idx,
-                                    'j': vec_idx,
-                                    'offset': self.c_offset_name(i, 0),
-                                    'dim': d.dataset.cdim})
+                        lhs = ast.Symbol(self.c_vec_name(), rank=(vec_idx,))
+                        rhs = ast.Symbol(m.offset[idx] * d.dataset.cdim)
+                        # rhs = ast.Symbol(self.c_offset_name(i, 0), rank=(idx,))
+                        # rhs = ast.Prod(rhs, ast.Symbol(d.dataset.cdim))
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -326,31 +399,43 @@ def c_zero_tmp(self, i, j):
...
+                        val_ast.append(ast.Incr(lhs, rhs))
+                        # val.append("%(name)s[%(j)d] += %(off)d * %(dim)s;" %
+                        #            {'name': self.c_vec_name(),
+                        #             'i': idx,
+                        #             'j': vec_idx,
+                        #             'offset': self.c_offset_name(i, 0),
+                        #             'dim': d.dataset.cdim,
+                        #             'off': m.offset[idx]})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -403,56 +488,84 @@ def c_map_decl(self, is_facet=False):
...
+                                rhs = ast.Sum(rhs, ast.Symbol(k))
+                            val_ast.append(ast.Assign(lhs, rhs))
+                            # val.append("xtr_%(name)s[%(ind_flat)s] = %(dat_dim)s * (*(%(name)s + i * %(dim)s + %(ind)s)%(off_top)s)%(offset)s;" %
+                            #            {'name': self.c_map_name(i, j),
+                            #             'dim': m.arity,
+                            #             'ind': idx,
+                            #             'dat_dim': d.cdim,
+                            #             'ind_flat': m.arity * k + idx,
+                            #             'offset': ' + '+str(k) if k > 0 else '',
+                            #             'off_top': ' + start_layer * '+str(m.offset[idx]) if is_top else ''})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -403,56 +488,84 @@ def c_map_decl(self, is_facet=False):
...
+                        lhs = ast.Symbol("xtr_" + self.c_map_name(i, j), rank=(idx,))
+                        rhs = ast.Symbol(self.c_map_name(i, j), rank=(ast.Sum(ast.Prod(ast.Symbol("i"), ast.Symbol(m.arity)), ast.Symbol(idx)),))
+                        if is_top:
+                            rhs = ast.Sum(rhs, ast.Prod(ast.Symbol("start_layer"), ast.Symbol(m.offset[idx])))
+                        val_ast.append(ast.Assign(lhs, rhs))
+                        # val.append("xtr_%(name)s[%(ind)s] = *(%(name)s + i * %(dim)s + %(ind)s)%(off_top)s;" %
+                        #            {'name': self.c_map_name(i, j),
+                        #             'dim': m.arity,
+                        #             'ind': idx,
+                        #             'off_top': ' + start_layer * '+str(m.offset[idx]) if is_top else ''})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -403,56 +488,84 @@ def c_map_decl(self, is_facet=False):
...
+                                    rhs = ast.Sum(rhs, ast.Symbol(k))
+                                val_ast.append(ast.Assign(lhs, rhs))
+                                # val.append("xtr_%(name)s[%(ind_flat)s] = %(dat_dim)s * (*(%(name)s + i * %(dim)s + %(ind)s)%(off)s)%(offset)s;" %
+                                #            {'name': self.c_map_name(i, j),
+                                #             'dim': m.arity,
+                                #             'ind': idx,
+                                #             'dat_dim': d.cdim,
+                                #             'ind_flat': m.arity * (k + d.cdim) + idx,
+                                #             'offset': ' + '+str(k) if k > 0 else '',
+                                #             'off': ' + ' + str(m.offset[idx])})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -403,56 +488,84 @@ def c_map_decl(self, is_facet=False):
...
+                                rhs = ast.Sum(rhs, ast.Symbol("start_layer"))
+                            rhs = ast.Sum(rhs, ast.Symbol(m.offset[idx]))
+                            val_ast.append(ast.Assign(lhs, rhs))
+                            # val.append("xtr_%(name)s[%(ind)s] = *(%(name)s + i * %(dim)s + %(ind_zero)s)%(off_top)s%(off)s;" %
+                            #            {'name': self.c_map_name(i, j),
+                            #             'dim': m.arity,
+                            #             'ind': idx + m.arity,
+                            #             'ind_zero': idx,
+                            #             'off_top': ' + start_layer' if is_top else '',
+                            #             'off': ' + ' + str(m.offset[idx])})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -469,17 +582,22 @@ def c_map_bcs(self, sign):
...
+                        lhs = ast.Symbol("xtr_" + self.c_map_name(i, j), rank=(idx,))
+                        rhs = ast.Symbol(max_int)
+                        if sign == '-':
+                            rhs = ast.Neg(rhs)
+                        val_ast.append(ast.Incr(lhs, rhs))
+                        # val.append("xtr_%(name)s[%(ind)s] %(sign)s= %(val)s;" %
+                        #            {'name': self.c_map_name(i, j),
+                        #             'val': max_int,
+                        #             'ind': idx,
+                        #             'sign': sign})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -490,22 +608,27 @@ def c_map_bcs(self, sign):
...
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -513,34 +636,46 @@ def c_add_offset_map(self, is_facet=False):
...
-                                        'dim': d.cdim})
+                            lhs = ast.Symbol("xtr_" + self.c_map_name(i, j), rank=(m.arity * k + idx,))
+                            rhs = ast.Symbol(m.offset[idx] * d.cdim)
+                            val_ast.append(ast.Incr(lhs, rhs))
+                            # val.append("xtr_%(name)s[%(ind_flat)s] += %(off)s[%(ind)s] * %(dim)s;" %
+                            #            {'name': self.c_map_name(i, j),
+                            #             'off': self.c_offset_name(i, j),
+                            #             'ind': idx,
+                            #             'ind_flat': m.arity * k + idx,
+                            #             'dim': d.cdim})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -513,34 +636,46 @@ def c_add_offset_map(self, is_facet=False):
...
-                                   {'name': self.c_map_name(i, j),
-                                    'off': self.c_offset_name(i, j),
-                                    'ind': idx})
+                        lhs = ast.Symbol("xtr_" + self.c_map_name(i, j), rank=(idx,))
+                        rhs = ast.Symbol(m.offset[idx])
+                        val_ast.append(ast.Incr(lhs, rhs))
+                        # val.append("xtr_%(name)s[%(ind)s] += %(off)s[%(ind)s];" %
+                        #            {'name': self.c_map_name(i, j),
+                        #             'off': self.c_offset_name(i, j),
+                        #             'ind': idx})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -513,34 +636,46 @@ def c_add_offset_map(self, is_facet=False):
...
-                                            'dim': d.cdim})
+                                lhs = ast.Symbol("xtr_" + self.c_map_name(i, j), rank=(m.arity * (k + d.cdim) + idx,))
+                                rhs = ast.Symbol(m.offset[idx] * d.cdim)
+                                val_ast.append(ast.Incr(lhs, rhs))
+                                # val.append("xtr_%(name)s[%(ind_flat)s] += %(off)s[%(ind)s] * %(dim)s;" %
+                                #            {'name': self.c_map_name(i, j),
+                                #             'off': self.c_offset_name(i, j),
+                                #             'ind': idx,
+                                #             'ind_flat': m.arity * (k + d.cdim) + idx,
+                                #             'dim': d.cdim})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -513,34 +636,46 @@ def c_add_offset_map(self, is_facet=False):
...
-                                        'ind_zero': idx})
-        return '\n'.join(val)+'\n'
+                            lhs = ast.Symbol("xtr_" + self.c_map_name(i, j), rank=(m.arity + idx,))
+                            rhs = ast.Symbol(m.offset[idx])
+                            val_ast.append(ast.Incr(lhs, rhs))
+                            # val.append("xtr_%(name)s[%(ind)s] += %(off)s[%(ind_zero)s];" %
+                            #            {'name': self.c_map_name(i, j),
+                            #             'off': self.c_offset_name(i, j),
+                            #             'ind': m.arity + idx,
+                            #             'ind_zero': idx})
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -554,41 +689,66 @@ def c_offset_init(self):
...
+            if j:
+                rhs = ast.Sum(rhs, ast.Symbol(j))
+            rhs = ast.Deref(rhs)
+            val_ast.append(ast.Assign(lhs, rhs))
+        return val_ast
+        # return ";\n".join(["%(name)s[i_0*%(dim)d%(ofs)s] = *(%(ind)s%(ofs)s);\n" %
+        #                    {"name": buf_name,
+        #                     "dim": dim,
+        #                     "ind": self.c_kernel_arg(idx),
+        #                     "ofs": " + %s" % j if j else ""} for j in range(dim)])
```
kynan: Please remove commented code

pyop2/host.py
```diff
@@ -554,41 +689,66 @@ def c_offset_init(self):
...
+                val_ast.append(ast.Incr(lhs, rhs))
+        return val_ast
+        # return ";\n".join(["*(%(ind)s%(nfofs)s) %(op)s %(name)s[i_0*%(dim)d%(nfofs)s%(mxofs)s]" %
+        #                    {"ind": self.c_kernel_arg(count, i, j),
+        #                     "op": "=" if self.access == WRITE else "+=",
+        #                     "name": buf_name,
+        #                     "dim": dim,
+        #                     "nfofs": " + %d" % o if o else "",
+        #                     "mxofs": " + %d" % (mxofs[0] * dim) if mxofs else ""}
+        #                    for o in range(dim)])
```
kynan: Please remove commented code

kynan:
I don't have a particularly good suggestion how, but this seems to be an occasion to refactor the host code gen. With these changes the code gen becomes even more opaque.

pyop2/host.py
doru1004: I've left this and the other comments in on purpose because it's easier to understand what gets constructed using the AST. If I take this out I would have to replace it with a block of text explaining this in words. This would take longer and it would be less clear what the output should be.

pyop2/host.py
doru1004: There are couple of comments here that need removing.

pyop2/host.py
doru1004: Yes, these ones!

pyop2/host.py
wence-: No.  There is absolutely zero guarantee that any subsequent edits to this code will also update the comment accordingly.  COMMENTS ARE NOT DOCUMENTATION.

doru1004: You're right. How do you suggest we document this then?
For me it makes no difference as I know what the code does so it's very easy to remove this and not replace it with anything else. I would like though to leave something here that will allow people that are no familiar with the code to still be able to understand what's going on.

doru1004: Any thoughts on this?

wence-: You could, I don't know, document what it is the function is doing.  What do all the magic arguments mean, etc...

doru1004:
I don't this is the appropriate PR you want to refactor this code in. The goal of this PR is to replace like for like and use AST representation instead of the templated representation.

pyop2/host.py
miklos1: "COMMENTS ARE NOT DOCUMENTATION." => Yes. Tests are documentation, since they must be updated if they become outdated.

kynan:
I know. I'm just concerned that this is another rather big change that further cements the structure of the current code gen.

doru1004:
I don't quite understand what your point is. I don't see how this change is not on the path for making host.py a better place.

kynan:
It sure is, but it's only a step.

wence-:
NAK.  This is out of date and not being worked on, and takes this code in the wrong direction.

PR #439, opened 2015-04-27
Replace maybe_set_dat_dirty on GPU backends

chromy:
A refactoring (OP2/PyOP2@7657a4be3343255cd45ead4a70b99387d6be6f18) changed
maybe_set_dat_dirty to update_arg_data_state this updates the name in
for cuda.py and opencl.py


wence-:
I think that you can just remove these calls completely, the GPU backends call the base ParLoop compute function which calls update_arg_data_state anyway.

chromy:
Okay, I updated the patch to remove the calls.

kynan:
Looks sensible to me.

PR #443, opened 2015-04-30
(Re)-add ParLoop Kernel timer in sequential backend

chromy:
Is it okay to put this back? It was removed in 16fa81178fb6e8209f28bcc7091b30b6f4008737 but I can't tell if it was deliberate or not, the other backends still seem to have this timer (e.g. https://github.com/OP2/PyOP2/blob/master/pyop2/opencl.py#L552).


wence-:
Thanks, done.

PR #444, opened 2015-05-02
sparsity: temporarily make map buffers writeable

wence-:
Memoryviews require that the buffer be writeable, so do that in
build_sparsity and fill_with_zeros.


kynan:
I can believe that.

PR #445, opened 2015-05-07
base: Use new FindInstances tree visitor

wence-:


wence-:
PyOP2 modifications to land with coneoproject/COFFEE#44

PR #446, opened 2015-05-12
sparsity: Fix buffer overrun in fill_with_zeros

wence-:
We were using an array of values of size 2*X but reading 4*X values in the case of interior horizontal facets.  Fix that to allocate the array with size 4*X.


kynan:
Looks sane.

PR #447, opened 2015-05-14
VFS / interior facet / extruded fix for map generation

dorugeber:
This fixes the "advection weirdness on extruded meshes" bug.  Turns out the extruded maps were incorrectly flattening the VFS/interior facet combination.  Credit to Doru for the fix.

Tests in firedrakeproject/firedrake#515


wence-:
I think I buy this.  I think I also buy why it's only needed for Dats (no change from before).

dorugeber:
FWIW I have no idea why it's "only needed for Dats", so I hope you're super-convinced about that!

wence-:
Yeah, hence the request on the firedrake side to test that (in some sense).

PR #448, opened 2015-05-19
Add option to check if code differs across processes

wence-:
If configuration["check_src_hashes"] is True (or, as before
configuration["debug"] is True) then verify that all processes are
indeed trying to compile the same code.


miklos1:
Fine.

PR #450, opened 2015-07-10
BC components

wence-:
PyOP2 side of the changes to allow setting BCs on components of vector-valued dats.


kynan:
I think I can believe this.

PR #451, opened 2015-07-16
Assembly cache fix

wence-:
Addresses a number of long-running assembly cache bugs in Firedrake.  In particular, appears to fix firedrakeproject/firedrake#319 firedrakeproject/firedrake#386 and firedrakeproject/firedrake#541.

However, I'm still not sure these are all correct fixes.  I basically do not understand the interaction of all the pieces fully.


kynan:
I don't understand those interactions either, but your changes look sane I think.

PR #452, opened 2015-07-22
Prefork

wence-:
Use Andreas Kloeckner's prefork package to avoid problems with fork and MPI.

Since the prefork part of pytools is so small, I have incorporated it wholesale (it's MIT licensed).  If we'd rather not do that I can do the dependency dance.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -53,7 +53,6 @@ from utils import *
 from backends import _make_object
 from mpi import MPI, _MPI, _check_comm, collective
 from profiling import timed_region, timed_function
-from sparsity import build_sparsity
 from version import __version__ as version
 
 from coffee.base import Node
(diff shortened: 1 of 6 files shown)


miklos1:
There's a minor outstanding linting problem which makes Travis unhappy.

This patch looks good, and it reduces branches in the code and also removes a configuration option which now becomes obsolete (?).
Does prefork always happen, and does it happen on all MPI processes? Does it mean that from now on we will have twice the amount processes whether we need it or not? Could that have a harmful impact of performance / resource usage?

I'm unsure whether it's better to include prefork or to add it as dependency.

wence-:
There's no way not to fork on all processes, since I can't check the rank before mpi is initialised. I guess one could kill the children on non rank 0, after mpi comes up. I hope the OS just sleeps the processes so it shouldn't be too much of a performance problem.

kynan:
These changes make sense to me. Given that it's only 1 module I won't object to importing it wholesale. Once Travis is happy I think this is good to go.

PR #454, opened 2015-07-24
Add Mat.set_local_diagonal_entries

wence-:
Remove Mat.inc_local_diagonal_entries since we were always intending to
use it for setting a diagonal.

PyOP2 half of addressing firedrakeproject/firedrake#558.


kynan:
Makes sense to me.

PR #457, opened 2015-08-05
Log flops

wence-:
Floating point logging for parallel loops (via PETSc for host backends)


FabioLuporini:
I can believe this

kynan:
LGTM.

PR #458, opened 2015-08-07
Fix kernel caching

FabioLuporini:
Avoid transforming an AST through COFFEE on the first kernel cache hit, unless loop fusion is enabled


kynan:
Makes sense to me.

PR #460, opened 2015-08-21
Install script

dham:
This splits up the requirements files in a manner convenient for the Firedrake install script. It's pretty much trivial.


miklos1:
Fine.

PR #415, opened 2015-09-25
No sane way to set compiler flags as a user.

kynan:
Patrick (rightly) comments that there's no sane way to modify the default set of compiler flags.  Either from pyop2 directly or through firedrake.  There should be.

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -60,8 +60,10 @@ class Compiler(object):
     :arg ld: Linker executable (optional, if ``None``, we assume the compiler
         can build object files and link in a single invocation, can be
         overridden by exporting the environment variable ``LDSHARED``).
-    :arg cppargs: A list of arguments to the C compiler (optional).
-    :arg ldargs: A list of arguments to the linker (optional).
+    :arg cppargs: A list of arguments to the C compiler (optional, prepended to
+        any flags specified as the cflags configuration option)
+    :arg ldargs: A list of arguments to the linker (optional, prepended to any
+        flags specified as the ldflags configuration option).
     :arg cpp: Should we try and use the C++ compiler instead of the C
         compiler?.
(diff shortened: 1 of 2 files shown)


kynan:
Any thoughts on this @wence- @dham?

FabioLuporini:
This looks simple and effective, so I'm OK with it. It remains the fact that we still have different places in which compiler options are added (the user can specify them; coffee adds flags; pyop2 comes with some default compiler options; and perhaps some other extra env variable?), which might be a little confusing in some scenarios (is it possible that one adds a compiler flag in one place, but this gets actually ignored because another flag coming later in the compile line overwrites it?).

Anyway, as I said, I'm totally fine with this, especially as short/middle term solution

kynan:
Yes, this is certainly only a first step, not claiming to be addressing all issues mentioned.

dham:
Looks OK but is there some way of testing this?

kynan:
Possibly. We could probably at least test the part that sets flags based on the environment variables.

wence-:
NAK to this change.  I want fewer magic environment variables.  Not more.  My comments on the interaction between coffee and PyOP2 still stand.  The issue is unresolved, so I'd like to leave this open.  But without keeping the pull request open because I don't want this change.

dham:
OK. I want to reconsider merging this. The specific use case which I think it covers is the following: wierd broken platform where gcc needs special flag combinations for JIT to work. I think that new environment variables are the Right Thing in this case. The suggested Python solutions break the portability of the user code and are IMHO the Wrong Thing for this use case. @wence- ?

wence-:
I think I buy this rationale.  This does give you a way to set flags from firedrake, by poking `parameters["pyop2_options"]["cflags"]`.  My concerns about the interaction with coffee and so forth are, I believe, still valid, but plausibly not relevant to the particular point here.

PR #465, opened 2015-10-02
Cpp kernels

wence-:
Support specifying C++ as kernel language.  Also switches some default optimisation options.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,12 +14,11 @@ env:
     - PETSC_CONFIGURE_OPTIONS="--download-ctetgen --download-triangle --download-chaco"
 # command to install dependencies
 before_install:
-  - sudo add-apt-repository -y ppa:cython-dev/master-ppa
   - sudo apt-get update -qq
   - "sudo apt-get install -qq build-essential python-dev git-core mercurial \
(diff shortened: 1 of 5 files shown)


.travis.yml
```diff
@@ -14,12 +14,11 @@ env:
     - PETSC_CONFIGURE_OPTIONS="--download-ctetgen --download-triangle --download-chaco"
 # command to install dependencies
 before_install:
-  - sudo add-apt-repository -y ppa:cython-dev/master-ppa
   - sudo apt-get update -qq
   - "sudo apt-get install -qq build-essential python-dev git-core mercurial \
     cmake cmake-curses-gui libmed1 gmsh python-pip libhdf5-openmpi-dev \
     libopenmpi-dev openmpi-bin libblas-dev liblapack-dev gfortran \
-    triangle-bin cython"
```
kynan: Why do we no longer need Cython?

kynan:
Looks good!

.travis.yml
wence-: We do need it.  The commit that removes this line says "Don't install cython from ppa".  Since it seems that the cython ppa breaks on travis, whereas the pip installed cython (that happens later) doesn't

kynan: OK, makes sense. Sorry, hadn't read the commit message.

PR #466, opened 2015-10-02
Multiple top bottom masks

dham:
Enable multiple different masks in order to support geometric boundary conditions on extruded domains.

### DIFF
--- a/pyop2/host.py
+++ b/pyop2/host.py
@@ -538,7 +538,7 @@ for ( int i = 0; i < %(dim)s; i++ ) %(combine)s;
                 for location, name in m.implicit_bcs:
                     if location == "bottom":
                         if bottom_masks is None:
-                            bottom_masks = m.bottom_mask[name]
+                            bottom_masks = m.bottom_mask[name].copy()
                         else:
                             bottom_masks += m.bottom_mask[name]
                         need_bottom = True
@@ -565,7 +565,7 @@ for ( int i = 0; i < %(dim)s; i++ ) %(combine)s;
                 for location, name in m.implicit_bcs:
                     if location == "top":
                         if top_masks is None:
-                            top_masks = m.top_mask[name]
+                            top_masks = m.top_mask[name].copy()
                         else:
                             top_masks += m.top_mask[name]
                         need_top = True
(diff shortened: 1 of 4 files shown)


miklos1:
I think this is OK, except that it seems to break the PyOP2 test suite.

wence-:
I think https://github.com/OP2/PyOP2/commit/d4eb085d977035eb8d134955d58e11a5cdeb7633 needs merging into master

pyop2/host.py
```diff
@@ -538,16 +538,22 @@ def c_map_bcs(self, sign):
...
-                    continue
-                need_bottom = True
-                for idx in range(m.arity):
-                    if m.bottom_mask[idx] < 0:
-                        val.append("xtr_%(name)s[%(ind)s] %(sign)s= %(val)s;" %
-                                   {'name': self.c_map_name(i, j),
-                                    'val': max_int,
-                                    'ind': idx,
-                                    'sign': sign})
+                bottom_masks = None
```
kynan: Wouldn't it make more sense for this to start as an empty list?

miklos1: I agree.

pyop2/host.py
```diff
@@ -559,18 +565,25 @@ def c_map_bcs(self, sign):
...
-                    continue
-                need_top = True
-                for idx in range(m.arity):
-                    if m.top_mask[idx] < 0:
-                        val.append("xtr_%(name)s[%(ind)s] %(sign)s= %(val)s;" %
-                                   {'name': self.c_map_name(i, j),
-                                    'val': max_int,
-                                    'ind': idx,
-                                    'sign': sign})
+                top_masks = None
```
miklos1: Same as above.

miklos1:
I think this is mostly okay, although GitHub says there is a merge conflict.

pyop2/host.py
dham: I disagree. Bottom_masks is never a list. When it's not None it's a numpy array.

miklos1: Oh, I didn't realise that. In that case I think `np.zeros` could be a sensible initial value.

dham: You could do that, but then the "if bottom_masks ..." below becomes a requirement to check that every entry in the mask is zero. I'm not convinced that's much of an improvement overall.

miklos1: I wonder if the current version might potentially be dangerous. Line 541 seems to copy a reference to a numpy array, while 543 seems to modify that.

dham: Ouch! Nice catch.

PR #467, opened 2015-10-07
Fix parallel compilation

miklos1:
Running the Firedrake test suite takes a lot of time. Running them in parallel makes in faster, but annoyingly a few random failures always appear when running the test suite in parallel with a clean cache. This pull request - I claim - fixes that problem.

Implementation ideas:
 * Make sure that no two processes writes the same file at a time by adding the process ID to the file name.
 * To still share files between processes, we rename the final file to have a process-independent name. This should be safe, as POSIX requires file moves to be atomic.

### DIFF
--- a/pyop2/caching.py
+++ b/pyop2/caching.py
@@ -288,8 +288,12 @@ class DiskCached(Cached):
         c = MPI.comm
         # Only rank 0 stores on disk
         if c.rank == 0:
+            # Concurrently writing a file is unsafe,
+            # but moving shall be atomic.
             filepath = os.path.join(cls._cachedir, key)
+            tempfile = os.path.join(cls._cachedir, "%s_p%d.tmp" % (key, os.getpid()))
             # No need for a barrier after this, since non root
             # processes will never race on this file.
-            with gzip.open(filepath, 'wb') as f:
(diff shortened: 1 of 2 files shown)


miklos1:
Maybe I should use `tempfile.mkstemp` instead of putting the process ID in the name.

kynan:
Agreed, `tempfile.mkstemp` would be better.

miklos1:
Okay, so I was thinking more about `tempfile.mkstemp`, and it can be applied to:

 * Saving the pickled `FFCKernel` object (good)
 * Writing the `.c` file
 * Writing the `.log` and `.err` files (irrelevant: I won't be reading those when running tests in parallel)

However, regarding the output of the compiler (`.o` and `.so` files), we cannot do any better than trying to guess a name that no one else would guess. So we cannot achieve `mkstemp`-level safety in the overall pipeline.

I would argue to leave this patch as it is and merge it. Currently the problem arises when to different Python programs (not two different MPI ranks of the same program) try to write the same file at the same time. With this patch, these processes must also have the same PID for conflict. That is not possible on a single computer, only when the cache directory is shared between different computers (why would one do that?) which are running Python with the same PID (unlikely) which are trying to run write the same file at the same time (unlikely).

PR #468, opened 2015-10-07
Hardcode offsets in semi-structured wrappers

miklos1:
This change hardcodes the offset values of the structured dimension in the wrapper.

@wence-: Please have a look at f26939b. I use the whole map in the key instead of just replacing arity with the offset array. This might be too discriminating? However, if it is equivalent to adding the individual components by hand, than I prefer this simpler version.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -4228,19 +4228,6 @@ class ParLoop(LazyComputation):
         return [arg for arg in self.args if arg._is_global_reduction]
 
     @cached_property
-    def offset_args(self):
-        """The offset args that need to be added to the argument list."""
-        _args = []
-        for arg in self.args:
-            if arg._is_indirect or arg._is_mat:
-                maps = as_tuple(arg.map, Map)
-                for map in maps:
-                    for m in map:
-                        if m.iterset._extruded:
-                            _args.append(m.offset)
-        return _args
-
-    @cached_property
     def layer_arg(self):
         """The layer arg that needs to be added to the argument list."""
         if self._is_layered:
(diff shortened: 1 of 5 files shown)


wence-:
Please don't use the whole map.  This will completely defeat the garbage collector.

wence-:
I think we're careful to not use anything in the cache key that results in holding on to things they reference (but maybe the implicit_bcs and or vector_index already do so)

pyop2/utils.py
```diff
@@ -68,7 +68,7 @@ def __get__(self, obj, cls):
 def as_tuple(item, type=None, length=None):
     # Empty list if we get passed None
     if item is None:
-        t = []
+        t = ()
```
kynan: oops...

miklos1:
@wence-: Fixed caching.

Still waiting for Travis, but otherwise I think this is ready to go.

pyop2/base.py
```diff
@@ -3856,24 +3856,25 @@ def _cache_key(cls, kernel, itspace, *args, **kwargs):
...
                 else:
                     idx = arg.idx
-                map_arity = arg.map.arity if arg.map else None
+                map_offset = tuplify(arg.map.offset) if arg.map else None
                 if arg._is_dat_view:
                     view_idx = arg.data.index
                 else:
                     view_idx = None
-                key += (arg.data.dim, arg.data.dtype, map_arity,
+                key += (arg.data.dim, arg.data.dtype, map_offset,
```
wence-: If the map is not extruded, and therefore `arg.map.offset` is `None` this key no longer treats maps with different arities as different.

miklos1: Ugh, right. Sadly nothing tests this in the Firedrake test suite. I'll fix.

PR #472, opened 2015-10-15
Refactor wrapper for point evaluations

miklos1:
This pull request tries to somewhat decouple extraction of cell-wise data and the iteration loop in the wrapper. The idea is not to re-implement (and then maintain at two places) the logic about extracting cell-wise data from the global arrays.

While I believe this change isn't very nice, it is part of a set of changes which will deliver a much needed feature for our users, and properly "fixing" `host.py` would delay that a lot.


miklos1:
@dham, @wence-: Could someone review this, please?

wence-:
I think this looks fine too.

PR #473, opened 2015-11-06
Remove h5py requirement

wence-:
This is optional anyway (it's only used explicitly in the demos) and I need to be able to build h5py manually in firedrake-install.


miklos1:
Fine.

PR #474, opened 2015-11-13
Flush assembly

wence-:
Correctly calls Mat assembly with assembly type flush on petsc matrices when switching between adding and inserting values.

Switches the travis build to build PETSc with debugging enabled to try and catch these kind of issues.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3539,6 +3539,35 @@ class Sparsity(ObjectCached):
         return False
 
 
+class _LazyMatOp(LazyComputation):
+    """A lazily evaluated operation on a :class:`Mat`
+
+    :arg mat: The :class:`Mat` this operation touches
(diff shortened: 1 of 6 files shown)


pyop2/base.py
```diff
@@ -3539,6 +3539,35 @@ def __contains__(self, other):
...
+class _LazyMatOp(LazyComputation):
+    """A lazily evaluated operation on a :class:`Mat`
+
+    :arg mat: The :class:`Mat` this operation touches
+    :arg closure: a callable piece of code to run
+    :arg new_state: What is the assembly state of the matrix after running
+         the closure?
+    :kwarg read:  Does this operation have read semantics?
+    :kwarg write:  Does this operation have write semantics?
+    :kwarg state: The state of the matrix after calling ``closure``.
```
kynan: This `kwarg` doesn't seem to exist?

PR #480, opened 2016-02-05
Check kernel hashes

wence-:
Addresses firedrakeproject/firedrake#591.  I do this in the source compilation to catch possible errors in the pyop2 code pasting stage.  I'd rather not put the check in the par_loop constructor, since that is called much more frequently.

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -33,6 +33,7 @@
 
 import os
 from mpi import MPI, collective
+from mpi4py import MPI as _MPI
 import subprocess
 import sys
 import ctypes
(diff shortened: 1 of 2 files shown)


miklos1:
Oh good, this implements my proposed idea. Have you tested its scalability on many nodes?

wence-:
No, I tried it on my laptop.  It should be fine, because it's just an allreduce, and those are claimed scalable by the MPI implementation.  We have reductions in many other places and haven't run into issues.  It should be no worse than the barriers in the compilation phase that are required to communicate via side-channel through the filesystem.

PR #482, opened 2016-02-19
Improve performance of the gen code for scatter

FabioLuporini:
With the contribution of @doru1004 and with extensive use of VTune, it appears that the performance of parloops writing and/or incrementing data indirectly is on average more than 10% faster with this new code generation scheme.

I checked this with both ICC and GCC 

I believe the speed up is due to a number of things, including: vectorisability of the address calculation; avoiding recalculating the same address; partial unrolling of the loop, which may indirectly suggest the compiler the absence of conflicts between consecutive writes; etc. Honestly, I don't have *the* explanation, which appears to be too fine grained even for VTune

as far as I can tell, it passes all PyOP2 tests and also all firedrake tests

feel free to run some regression tests

Ah, the code is also slightly more readable (...) now (at least, I've explicitly listed all possible cases in which a buffer may be generated)

### DIFF
--- a/pyop2/host.py
+++ b/pyop2/host.py
@@ -648,13 +648,40 @@ for ( int i = 0; i < %(dim)s; i++ ) %(combine)s;
                             "ofs": " + %s" % j if j else ""} for j in range(dim)])
 
     def c_buffer_scatter_vec(self, count, i, j, mxofs, buf_name):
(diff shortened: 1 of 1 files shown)


wence-:
Great! Merging.

wence-:
Oh, wait, some of the firedrake tests fail for me.

wence-:
I have fixed that problem.

PR #487, opened 2016-04-22
Fix monolithic partially mixed matrices

wence-:
If we have a matrix that is mixed in one direction but uses a plain
dataset in the other direction we previously would not be able to build
it if the dataset had cdim != 1 because block sizes didn't match.  Fix
this by providing an unblocked_lgmap property on the DataSets and using
it appropriately when building monolithic mixed matrices.


miklos1:
Looks fine.

PR #488, opened 2016-05-09
Petsc logging

wence-:


.travis.yml
```diff
@@ -37,6 +37,7 @@ before_install:
        --allow-external petsc --allow-unverified petsc \
        --allow-external petsc4py  --allow-unverified petsc4py \
        < requirements-git.txt"
+  - pip install pulp
```
miklos1: If you install PuLP, you probably also want to add `glpk-utils` at the apt dependencies.

miklos1:
Looks good.

.travis.yml
wence-: I only need this so that the pyop2 tests pass on travis (i.e. coffee is importable).

PR #489, opened 2016-05-24
Communicators

wence-:
Allow building PyOP2 objects on something other than COMM_WORLD.  I think I've caught them all!


pyop2/caching.py
```diff
@@ -234,66 +229,3 @@ def _cache_key(cls, *args, **kwargs):
     def cache_key(self):
         """Cache key."""
         return self._key
-
-
-class DiskCached(Cached):
```
miklos1: Why does this go?

setup.py
```diff
@@ -148,7 +148,7 @@ def run(self):
       install_requires=install_requires,
       dependency_links=dep_links,
       test_requires=test_requires,
-      packages=['pyop2', 'pyop2_utils'],
```
miklos1: Is this removed for the second time?

pyop2/caching.py
miklos1: Ah, I see this is merged into `TSFCKernel`, probably the only user of `DiskCached`.

wence-:
Comments on this, especially the implementation of communicator duplication welcome!

pyop2/mpi.py
```diff
@@ -33,9 +33,184 @@
...
+            dupped_comms.append(comm_out)
+        else:
+            refcount = comm_out.Get_attr(refcount_keyval)
+            if refcount is None:
+                raise ValueError("Inner comm without a refcount")
+            refcount[0] += 1
+    return comm_out
+
+
+def free_comm(comm, remove=True):
```
miklos1: Is this supposed to be called by the user? The only caller I see is `free_comms` at exit.

wence-: It is also used by firedrake.  Where the comm goes out of scope at the end of (e.g.) https://github.com/firedrakeproject/firedrake/blob/3b75264c23120488d6e8fc5faded21ef5375224f/firedrake/mesh.py#L273

miklos1: Ok, I see, so `dup_comm` and `free_comm` appear in pairs like with a context manager, and it provides "lexical scoping" for the communicator.

miklos1:
I think this looks mostly fine.

PR #493, opened 2016-07-06
Loop fusion engine

FabioLuporini:
*** PREMISE: I know the list of commits is awfully long, but thankfully the pyop2 code base is only marginally touched. Loop fusion lives almost entirely under pyop2/fusion. ***

This PR adds three types of loop fusion:

- "Soft fusion": kernels over the same iteration space not presenting indirect data dependencies are fused into a single kernel
- "Hard fusion": kernels over different iteration spaces characterised by an indirect INC-after-INC dependence (e.g., facet+cell integrals in finite element methods) are fused using a "special scheme". Sticking to the facet+cell integrals example, the idea is to execute one facet and then spin over the adjacent cells; the cells are also executed as long as it's the first time they are visited.
- "Sparse tiling": any two consecutive loops can be fused, irrespective of data dependencies (read-after-write, write-after-read, etc.). 

While soft fusion and hard fusion are applied automatically by PyOP2 (without any kind of user intervention), sparse tiling:
- has to be "requested" explicitly through the ``loop chain`` interface (a context manager to be placed around a piece of code within which parloops are generated, see example below)
- requires the SLOPE library (https://github.com/coneoproject/SLOPE) be available. This is because the data dependence pattern may be arbitrarily complex, and a special type of analysis at the "mesh-level" (i.e., in C land) is required. In essence: PyOP2 looks at the trace of loops, builds a representation suitable for SLOPE, and then uses the Python interface shipped by SLOPE to create two kernels, the ``inspector`` and the ``executor``. The inspector is JIT-compiled and executed; this creates a data structure -- the "inspection" -- that contains enough information for running the executor. The executor, which represents a fused sequence of kernels, is run every time the same loop chain is encountered. The PyOP2 caching system is exploited to avoid repeating inspection over and over.

An example:

If we have:
```
with loop_chain (...):
    parloop0
    parloop1
    parloop2
    parloop3
```

Then PyOP2 will first try to apply soft to parloop{0,1,2,3}; then hard fusion; finally, if not al loops got fused, it will try SLOPE. If SLOPE isn't available, this last step is skipped. If the sequence of loops is not wrapped in a loop chain context, then only soft and hard fusion will be attempted.

Some unit tests are provided by this PR. All firedrake tests pass. One simple way of appreciating loop fusion in Firedrake is running the tests and then running ``grep "fuse" *c`` or ``grep "Fuse" *c`` from the pyop2-cache directory. This will catch all kernels that have been subjected to hard (grep "fuse") and soft fusion (grep "Fuse").

Loop fusion is by default *DISABLED*. One can turn it on by setting the environment variable PYOP2_LOOP_FUSION=1 or by setting, at run-time, ``configuration[''loop_fusion"] = True``. The unit tests provided in PyOP2, however, enforce fusion irrespective of the value of this environment variable. The reason loop fusion is not ON by default is that I'd like to perform more extensive testing. I've stressed the fusion engine in many different ways and with many types of tests, but I'd prefer some beta testing before this goes live (nothing special -- just someone working with PYOP2_LOOP_FUSION=1 for a bunch of days). 

@wence- :  time ago I explained to you that hard fusion was suffering from an implementation flaw. To remind you and to make a long story short, consider the following example:
- I have one loop over facets accessing the coefficient ``f1``
- There is then a loop over cells that increments into the same global vector which accesses ``f1, f2, f3``
The implementation flaw was that all coefficients ``f1, f2, f3`` were loaded for all cells adjacent to a facet, no matter whether a cell had already been executed or not. When I realised that, I saw how awful and inefficient that was. Thanks to the latest refactoring, this problem has gone. The generated code is also "more efficient" by avoiding unnecessary gathers and scatters. However, I still do not have performance results: it would be of great help if we could work them out together, as I don't have much time in the near future. But the transformation is there, it works, so it'd be a pity if we didn't do this properly. Incidentally, this stuff doesn't appear in my thesis.

pyop2/fusion contains five modules.

- interface.py : this is self-explanatory. The key functions to trigger loop fusion/tiling are exposed here
- extended.py : this extends several class hierarchies in base.py. Special ``Arg``, ``Kernel``, ``ParLoop`` etc. are here defined. 
- filters.py : mostly utility functions to merge list of arguments belonging to different kernels (the intersection may be non-empty)
- transformer.py : the main module. The main class ``Inspector`` analyses a trace of loops and tries to apply loop fusion (soft, hard, tiling). Some form of data dependence analysis is performed. New ``Kernels`` are constructed (and cached). ASTs for the fused kernels are produced. 
- scheduler.py : schedulers use the information produced by an ``Inspector`` to create ``fusion.ParLoop`` objects. For example, given the trace of loops ``[L0, L1, L2, L3, L4, ..., Ln]``, a scheduler may fuse some of these loops and generate a new trace that could look like ``[L2', L3, L5', ..., Ln]``, where ``L2' = fusion(L0,L1,L2)``and ``L5'= fusion(L4, L5)`` (L3 is so untouched).


pyop2/base.py
```diff
@@ -165,11 +165,10 @@ def _depends_on(reads, writes, cont):
         self._trace = new_trace
 
         if configuration['loop_fusion']:
-            from fusion import fuse
-            to_run = fuse('from_trace', to_run, 0)
-        with timed_region("Trace: eval"):
-            for comp in to_run:
-                comp._run()
+            from fusion.interface import fuse, lazy_trace_name
```
wence-: I know we don't currently do it, but please use absolute imports.

pyop2/fusion/__init__.py
```diff
@@ -0,0 +1 @@
+from interface import loop_chain, loop_chain_tag  # noqa
```
wence-: absolute imports please

pyop2/fusion/extended.py
```diff
@@ -0,0 +1,706 @@
...
+from hashlib import md5
+
+import pyop2.base as base
+import pyop2.sequential as sequential
+import pyop2.host as host
+from pyop2.utils import flatten, strip, as_tuple
+from pyop2.mpi import collective
+from pyop2.profiling import timed_region
+
+from interface import slope, lazy_trace_name
```
wence-: likewise.

pyop2/fusion/extended.py
```diff
@@ -0,0 +1,706 @@
...
+            # A single AST (as a result of soft or hard fusion)
+            self._ast = fused_ast
+            self._code = self._ast_to_c(fused_ast)
+        else:
+            # Multiple functions (AST or strings, as a result of tiling)
+            self._ast = None
+            self._code = self._multiple_ast_to_c(kernels)
+        self._kernels = kernels
+
+        self._initialized = True
```
wence-: I am concerned that this guy inherits from both `sequential.Kernel` and `tuple` yet calls the constructor of neither.

pyop2/fusion/extended.py
```diff
@@ -0,0 +1,706 @@
...
+        self._it_space = kwargs['it_space']
+        super(FusionParLoop, self).__init__(kernel, iterset, *args, **kwargs)
+
+    def _build_itspace(self, iterset):
+        """
+        Bypass the construction of a new iteration space.
+
+        This avoids type checking in base.ParLoop._build_itspace, which would
+        return an error when the fused loop accesses arguments that are not
+        accessed by the base loop.
```
wence-: Huh?

pyop2/fusion/extended.py
```diff
@@ -0,0 +1,706 @@
...
+
+    """A simple bag of :class:`IterationSpace` objects for a sequence of tiled
+    parallel loops."""
+
+    def __init__(self, all_itspaces):
+        self._iterset = [i._iterset for i in all_itspaces]
+        self._extents = [i._extents for i in all_itspaces]
+        self._block_shape = [i._block_shape for i in all_itspaces]
+        assert all(all_itspaces[0].comm == i.comm for i in all_itspaces)
+        self.comm = all_itspaces[0].comm
```
wence-: Again, no superclass constructor called.

It looks like you're fighting really hard against the pyop2 interface to shoehorn stuff in.  We should note this going forward given any potential redesigns.

pyop2/fusion/interface.py
```diff
@@ -0,0 +1,331 @@
...
+
+from pyop2.base import _LazyMatOp
+from pyop2.mpi import MPI
+from pyop2.logger import warning, info as log_info
+from pyop2.utils import flatten
+
+try:
+    """Is SLOPE accessible ?"""
+    sys.path.append(os.path.join(os.environ['SLOPE_DIR'], 'python'))
+    import slope_python as slope
```
wence-: Don't do this.  Make slope installable as a python package.

pyop2/fusion/interface.py
```diff
@@ -0,0 +1,331 @@
...
+
+    # Set the SLOPE backend
+    backend = os.environ.get('SLOPE_BACKEND')
+    if backend not in ['SEQUENTIAL', 'OMP']:
+        backend = 'SEQUENTIAL'
+    if MPI.COMM_WORLD.size > 1:
```
wence-: In the new and brave world of pyop2 with communicators, this may not be right.

pyop2/fusion/interface.py
```diff
@@ -0,0 +1,331 @@
...
+    slope.set_exec_mode(backend)
+    log_info("SLOPE backend set to %s" % backend)
+except:
+    slope = None
+
+lazy_trace_name = 'lazy_trace'
+"""The default name for sequences of lazily evaluated :class:`ParLoop`s."""
+
+from transformer import Inspector
+import extended
```
wence-: absolute imports

pyop2/fusion/filters.py
```diff
@@ -0,0 +1,133 @@
...
+                        fa.access = RW
+                    elif WRITE in [a.access, fa.access]:
+                        # Can't be a READ, so just stick to WRITE regardless of what
+                        # the other access mode is
+                        fa.access = WRITE
+                    else:
+                        # Neither READ nor WRITE, so access modes are some
+                        # combinations of RW, INC, MIN, MAX. For simplicity,
+                        # just make it RW.
+                        fa.access = RW
```
wence-: I am concerned about the merging you're doing here.  Can you describe in what circumstances this happens and why you think it's safe.  For example, I don't think it's safe to convert `INC` into `RW`.

pyop2/base.py
FabioLuporini: OK

pyop2/fusion/__init__.py
FabioLuporini: Yes

pyop2/fusion/extended.py
FabioLuporini: OK

pyop2/fusion/extended.py
FabioLuporini: I don't get anything from invoking the ``sequential.Kernel``'s constructor as I'd have to replace all fields with the union of the information from the fused kernels

Actually this led me to a good spot, I think. I was using inheritance to regenerate the cache key of the kernels being fused, but this is already stored in the cache_key attribute... I think I can get rid of the ``super`` in ``_cache_key``, thus making the instantiation of a ``fusion.Kernel`` much faster. 

Note that inheritance is still necessary as I'm using some of the superclass' methods

pyop2/fusion/extended.py
FabioLuporini: When I fuse two kernels over different iteration spaces (e.g., in HardFusion), say X (base) and Y (fused) it may be that Y accesses fields that are not accessed by X. Now, Z = fuse(X, Y) has the same iteration space as X (base), but it might also have an indirect ``Arg`` from Y's iteration space to another set. In ``base.ParLoop``'s build_itspace we do type checking: we check that all maps from all indirect arguments have as ``fromset`` exactly X's iteration space -- which wouldn't be the case anymore here. 

Roughly speaking, a ``FusionParLoop`` relaxes the constraint that all indirect arguments should have the same base set (i.e., the iteration set of the loop)

pyop2/fusion/extended.py
FabioLuporini: Same reasons as before...

Oh yes, I fought really really hard. Especially for code generation

pyop2/fusion/interface.py
FabioLuporini: Mmm I see...what's the PyOP2 way of checking whether I'm running with more than one process ?

Previously I could use ``MPI.parallel``

pyop2/fusion/interface.py
FabioLuporini: Fixing

pyop2/fusion/filters.py
FabioLuporini: This I did in a previous PR. It should be safe, as ``RW`` is more restrictive than ``INC`` (you're simply losing associativity and commutativity). Here you've the same ``Dat`` accessed at least twice and only once with ``INC``: no matter what the other access does (surely not a ``READ``, surely not a ``WRITE``), but those two properties go lost, so the safest thing to do I can think of is to set it to ``RW``

FabioLuporini:
OK, pushed some fixes

FabioLuporini:
Now this depends firedrakeproject/firedrake#833

pyop2/fusion/interface.py
FabioLuporini: This has been fixed with the latest commit.

wence-:
Still some lint failures.  Otherwise I would like to merge this.  Is it still the case that kernel fusion (soft/hard) is off by default?  Would you be happy switching it on?

FabioLuporini:
uops, now flake8 shouldn't bark anymore

FabioLuporini:
I'm happy if we merge this. 

Yes, it's off by default, but what I propose we do is some "background" beta testing and then we switch it on. I.e., @wence-  and @miklos1 can work normally but with fusion ON in background. If you don't experience anything weird, then we can just switch it on after a few days. Are you happy with this?

wence-:
I would prefer to switch it on for everyone and announce on the mailing list.  That way we get broader test coverage.

FabioLuporini:
OK!

FabioLuporini:
I propose we do the following:
- we merge this
- we keep fusion disabled until I come back from holiday (early september)
- who feels brave enough, can start using/testing loop fusion as discussed above

miklos1:
I see Travis passes, but when I update an existing installation to this branch, I get on a non-developer install
```
======================================================= ERRORS =======================================================
_____________________________________ ERROR collecting test/unit/test_fusion.py ______________________________________
unit/test_fusion.py:42: in <module>
    import pyop2.fusion.interface
E   ImportError: No module named fusion.interface
================================== 14 passed, 1818 skipped, 1 error in 6.88 seconds ==================================
```

On a developer install I don't get previous error, but most tests are skipped (expect for 14, just as above).

FabioLuporini:
Ah yes ! We have to remove the old fusion.pyc on update

What's the best way of doing this?

miklos1:
I don't know what's going on. I'm more concerned about skipping all tests.

wence-:
Do you run `make test` or `py.test`.  If the latter you must do `py.test --backend=sequential ...`

miklos1:
I run `make test`. By most tests I mean `14 passed, 1818 skipped`.

miklos1:
On a different Firedrake installation this seems to work now.

FabioLuporini:
so can we merge it?

miklos1:
I guess, but let me test again. Remind me, this adds fusion as an option, but doesn't turn it on by default, is that correct?

miklos1:
OK, if I run the Firedrake test suite with loop fusion on, I get a few segfaults.

miklos1:
All those segfaults are in `tests/regression/test_multiple_domains.py`.

FabioLuporini:
Will take a look -- this test has probably been added recently, right ?

miklos1:
Yes, @wence- added that recently.

FabioLuporini:
@miklos1 : this gets fixed by merging https://github.com/coneoproject/COFFEE/pull/90

please let's do this ASAP.

miklos1:
This seems to work now. I guess we can merge this unless @wence- has any objections.

wence-:
I'm happy to hit go I think, we will no doubt find issues once it's live.  But let's test it for a wider user base.

PR #494, opened 2016-07-20
Only include diagonal in sparsity if row and column dataset are the same

stephankramer:
Currently entries are reserved for the diagonal in a sparsity regardless of whether the matrix is square or not. This leads to spurious nonzero entries in the PETSc matrices. This branch changes this to only reserve space for the diagonal when the row and column dataset are the same.

The main motivation for this is the performance of AMG preconditioners on the Schur complement of a mixed system. Take the case of the standard [[M, G], [C, 0]] saddlepoint system coming from du/dt+grad p=0 with div(u)=0. Although the extra nonzeros in the discretised gradient and continuity matrices G and C is relatively small, if you look at the 'selfp' fieldsplit preconditioner matrix for the Schur complement C M^-1 G the number of nonzeros is twice that what it needs to be. This leads to a dramatic performance degradation of the GAMG preconditioner (I've seen a factor of 10 slower for large systems) as the number of nonzeros at the coarser levels explodes. The performance of hypre+boomeramg is also improved by not including the diagonals in the G and C matrices (roughly twice, consistent with the n/o nonzeros).

More generally having a spurious diagonal for the non-square case may affect cache performance, as the column index of the diagonal will generally be wildly different than those of the other "local" column indices.

I presume the motivation for preallocating a diagonal in the *square* case is to ensure DirichletBCs can later be applied? Am I right in thinking this isn't necessary for the non-square case, as I'm not sure how DirichletBCs would work in that case? If this does need to be supported, can we make the allocation of a diagonal optional?

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3296,6 +3296,8 @@ class Sparsity(ObjectCached):
         self._nrows = self._rmaps[0].toset.size
         self._ncols = self._cmaps[0].toset.size
 
+        self._has_diagonal = self._rmaps[0].toset == self._cmaps[0].toset
+
         tmp = itertools.product([x.cdim for x in self._dsets[0]],
                                 [x.cdim for x in self._dsets[1]])
 
--- a/pyop2/petsc_base.py
+++ b/pyop2/petsc_base.py
@@ -376,6 +376,7 @@ class SparsityBlock(base.Sparsity):
         self._cmaps = tuple(m.split[j] for m in parent.cmaps)
         self._nrows = self._dsets[0].size
         self._ncols = self._dsets[1].size
+        self._has_diagonal = i == j and parent._has_diagonal
         self._parent = parent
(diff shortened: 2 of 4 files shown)


dorugeber:
@stephankramer I believe the thread Nullspace & VectorSpaceBasis in http://mailman.ic.ac.uk/pipermail/firedrake/2014-December/date.html (and again in http://mailman.ic.ac.uk/pipermail/firedrake/2015-April/date.html) led to preallocation of diagonal entries.  I think I agree that there's no need to do this for off-diagonal blocks.

wence-:
Effectively, a matrix always needs to have space allocated for the diagonal in petsc.  Now, you might think that this always happens.  However, if you make a form that only contains boundary integrals then not all of the dofs will be touched.  Hence the need to explicitly put the diagonal in.  I agree, I think that this is only necessary if the matrix is square (for application of bcs say).  So I think this change is good.  I am going to run the firedrake test suite with a debug petsc to be sure.

PR #497, opened 2016-08-16
Make debug bool instead of int

thomasfangyi:
`debug` should be a `bool` value instead of an `int`

This has caused problems in downstream when validating the type of this parameter. Since, the variable appears to be used as a `bool`, its type should be changed to `bool`.


wence-:
Thanks.

PR #498, opened 2016-08-23
Remove op2.Const

miklos1:
Reasons:
- Never used by Firedrake.
- Constants are globally defined, and every generated wrapper depends on all every constant ever defined which is terrible.
- On devices where `op2.Const` and `op2.Global` are different, the former is really just an optimisation for using the latter with READ access.


wence-:
ACK.

PR #500, opened 2016-09-19
Temporary fix for Mat buffer size when flatten=False

miklos1:
Fixes a bug that causes segmentation faults without `flatten=True`. FInAT does not use `flatten=True`.

### DIFF
--- a/pyop2/host.py
+++ b/pyop2/host.py
@@ -948,6 +948,10 @@ def wrapper_snippets(itspace, args,
             else:
                 _buf_size = [sum(_buf_size)]
                 _loop_size = _buf_size
+        else:
+            if not arg._flatten:
+                _dat_size = arg.data.dims[0][0]  # TODO: [0][0] ?
+                _buf_size = [e*d for e, d in zip(_buf_size, _dat_size)]
         _buf_decl[arg] = arg.c_buffer_decl(_buf_size, count, _buf_name[arg], is_facet=is_facet)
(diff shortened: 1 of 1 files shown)


pyop2/host.py
```diff
@@ -948,6 +948,10 @@ def extrusion_loop():
             else:
                 _buf_size = [sum(_buf_size)]
                 _loop_size = _buf_size
+        else:
+            if not arg._flatten:
+                _dat_size = arg.data.dims[0][0]  # TODO: [0][0] ?
```
wence-: The `dims` of a mat are a tuple of tuples.  Where the `[i][j]` entry says the row and column block size of the `i`th row and `j`th column.  `i` and `j` are not `0` if the matrix is mixed.  Here, you don't have any mixed matrices to contend with (hence 0, 0).  For example:

```python
V = FunctionSpace(mesh, "DG", 0)
Q = VectorFunctionSpace(mesh, "DG", 0)
W = V*Q*V

u = TrialFunction(V)
v = TestFunction(V)

dims = assemble(inner(u, v)*dx).M.dims
print dims
=> (((1, 1), (1, 2), (1, 1)), ((2, 1), (2, 2), (2, 1)), ((1, 1), (1, 2), (1, 1)))
```

miklos1: The Firedrake test suite never exercises the mixed matrix case.

PR #502, opened 2016-09-27
Allow for multiple fundecl objects in kernel AST

### LINKED ISSUE #501: Creating an OP2 Kernel with more than 1 FunDecl in AST
Hello,

I noticed a few changes to PyOP2 since I last used it. In particular, in `base.py`, line 3781, there is an assertion:
`asset len(fundecls) == 1, "Illegal Kernel"`

This is a problem for the specialized SLATE form compiler, as the AST I provide has more than a single FunDecl object: The body of the kernel has a macro function, and several sub-functions provided by TSFC for evaluating integral forms. As a result, I'm no longer able to produce the kernels I need. Since the changes.

My question is: Is there a particular reason for the assertion? And in particular, is there a way around this?

Thank you for any assistance!

thomasgibson:
This pull request adds a small change to the `_attached_info`  in the `Kernel` class of `base.py`, which allows kernels with multiple `fundecl` objects to be handled. This PR fixes #501.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3972,9 +3972,10 @@ class Kernel(Cached):
             self._code = self._ast_to_c(self._ast, opts)
             search = FindInstances(ast.FunDecl, ast.FlatBlock).visit(self._ast)
             fundecls, flatblocks = search[ast.FunDecl], search[ast.FlatBlock]
-            assert len(fundecls) == 1, "Illegal Kernel"
+            assert len(fundecls) >= 1, "Illegal Kernel"
+            fundecl, = [fd for fd in fundecls if fd.name == self._name]
             self._attached_info = {
-                'fundecl': fundecls[0],
+                'fundecl': fundecl,
                 'attached': False,
                 'flatblocks': len(flatblocks) > 0
             }


pyop2/base.py
```diff
@@ -3778,9 +3778,10 @@ def __init__(self, code, name, opts={}, include_dirs=[], headers=[],
             self._code = self._ast_to_c(self._ast, opts)
             search = FindInstances(ast.FunDecl, ast.FlatBlock).visit(self._ast)
             fundecls, flatblocks = search[ast.FunDecl], search[ast.FlatBlock]
-            assert len(fundecls) == 1, "Illegal Kernel"
+            assert len(fundecls) >= 1, "Illegal Kernel"
+            index = [i for i, fundecl in enumerate(fundecls) if fundecl.name == self._name][0]
```
miklos1: `[0]` at end is ugly, because if the condition is satisfied for more than one function declarations then the first is taken and the rest ignored, and I guess this case is not expected at all, so it is better to fail if it still does. Moreover, it is not really the index that you need later, but the function declaration itself, so I would write this line like this:
```python
fundecl, = [fd for fd in fundecls if fd.name == self._name]
```

thomasgibson: Ah, good call. I will update this change.

FabioLuporini:
Yes, we discussed about this problem and we agreed this was the way to go. 

Only one little thing. At the moment, loop fusion is by default off (my fault: I'm waiting for a little bit more free time so that if there were bugs I could fix them quickly), so could you also run the whole test suite (both Firedrake and PyOP2) with fusion ON, by simply doing ``export PYOP2_LOOP_FUSION=1`` just before running ``py.test ...`` ?

Thanks!

thomasgibson:
Hi @FabioLuporini, I just ran `export PYOP2_LOOP_FUSION=1` and then `make test`. Everything seemed to pass where it should. However, when I try `py.test test/` I get a large amount of skips. Is this normal?

miklos1:
@thomasgibson: Yes and no. When calling py.test directly in PyOP2, you ought to specify the backend with `--backend=sequential` (and similarly for the other backends: openmp, cuda, opencl), otherwise almost everything is skipped and hardly anything is being tested.

miklos1:
@thomasgibson: Moreover, you should also run the Firedrake test suite with the loop fusion environment variable, that will give you much better coverage.

thomasgibson:
Thanks, @miklos1! Yes I just ran `export PYOP2_LOOP_FUSION=1` and then `py.test test/ --backend=sequential` and much more was tested and passed. I do notice however with `test_fusion.py` I gets 8 passes and 15 skips. Should these 15 skipped tests become activated when loop fusion is activated?

miklos1:
`test_fusion.py` should enable loop fusion irrespective of the environment variable. Do you have SLOPE installed? You need that to do loop fusion.

thomasgibson:
I just installed SLOPE and ran the PyOP2 tests. It passes all tests in `test_fusion.py`. I will run the Firedrake test suite now to check everything is okay.

miklos1:
s/of/if/

thomasgibson:
Ok, so I'm getting some compilation errors in individual items when I run the test suite in Firedrake. Moreover, it still happens even when I'm switched to all master branches. Hold off on this PR until I can figure out if it's something wrong with my Firedrake installation, or something more.

thomasgibson:
Ok, so I managed to get a clean version working and I ran the Firedrake test suite with this branch checked out. Everything seems to work just fine.

FabioLuporini:
Perhaps you just had to clean up the cache?

thomasgibson:
Honestly, I'm not exactly sure what went wrong. It wasn't as simple as clearing the cache. I had multiple Firedrake packages installed and then I removed them, leaving one clean and working installation. That cleared up all the weird errors I was receiving.

miklos1:
Why is this not merged yet?

thomasgibson:
I can't merge this.



Sent from my Samsung Galaxy smartphone.


-------- Original message --------

miklos1:
With this branch, I'm getting a failure in a demo running the Firedrake test suite with loop fusion enabled. The generated code fails to compile:
```c
static inline void form0_interior_facet_integral_otherwise_form0_cell_integral_otherwise (double  A_0[8] , const double *const restrict *restrict coords__1 , const double *const restrict *restrict w_0_2 , const double *const restrict *restrict w_1_3 , const double *restrict w_2_4 , const unsigned int  facet_5[2] , const double *const restrict *restrict arg2_0 , int*  arg2_0_map0_0 , const double *restrict arg5_0 , int*  executed , int*  fused_iters , int  i )
{
  // Args aliases
  double * A_6  = A_0;
  const double *const restrict *restrict coords_7  = coords__1;
  const double *const restrict *restrict w_1_9  = w_0_2;
  const double *const restrict *restrict w_2_10  = w_1_3;
  const double *restrict w_4_12  = w_2_4;
  form0_interior_facet_integral_otherwise_base(A_0, coords__1, w_0_2, w_1_3, w_2_4, facet_5);
  i = fused_iters[0];
  if (!(executed[i])) {
    const double *const restrict w_1_3_vec[4] ;
    w_1_3_vec[0] = w_1_3[0];
    w_1_3_vec[1] = w_1_3[1];
    w_1_3_vec[2] = w_1_3[2];
    w_1_3_vec[3] = w_1_3[3];
```
There shouldn't be any `const` after the last star in `const double *const restrict w_1_3_vec[4];`

thomasgibson:
@miklos1 Did you check again per our discussion from earlier? The last time I ran the tests I didn't receive any failures. I'll check again for safety.

miklos1:
@thomasgibson: Yes, it still fails.

thomasgibson:
I am unable to reproduce the error. I ran the test suite twice and passed all tests (and yes I ran with all master branches, except for this one). Perhaps we should discuss this again in person and see what went wrong.

miklos1:
OK, so the problem is reproducible with PyOP2 master, so it is probably not caused by this patch. @FabioLuporini: the issue might be caused/triggered by firedrakeproject/tsfc#52.

PR #503, opened 2016-10-05
Globals matrices

dham:
This enables matrix blocks whose row and/or column set is a Global. The resulting blocks are dense and isomorphic to Dats or Globals. This is required in order to support the "Real" (or "Rognes") space in Firedrake.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3208,7 +3208,9 @@ class Sparsity(ObjectCached):
         self._dsets = dsets
 
         if dsets[0] is None or dsets[1] is None:
-            pass
+            # This will cause a trivial memory accounting error (although not a leak).
+            self._d_nz = 0
+            self._o_nz = 0
         else:
             self.lcomm = self._rmaps[0].comm
             self.rcomm = self._cmaps[0].comm
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3276,7 +3276,7 @@ class Sparsity(ObjectCached):
 
         # Check data sets are valid
         for dset in dsets:
-            if not isinstance(dset, DataSet):
+            if not isinstance(dset, DataSet) and dset is not None:
                 raise DataSetTypeError("All data sets must be of type DataSet, not type %r" % type(dset))
 
         # A single map becomes a pair of identical maps
@@ -3286,6 +3286,10 @@ class Sparsity(ObjectCached):
 
         # Check maps are sane
         for pair in maps:
+            if pair[0] is None or pair[1] is None:
+                # None of this checking makes sense if one of the
+                # matrix operands is a Global.
+                continue
             for m in pair:
                 if not isinstance(m, Map):
                     raise MapTypeError(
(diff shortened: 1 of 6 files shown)


pyop2/base.py
```diff
@@ -704,6 +702,75 @@ def fromhdf5(cls, f, name):
...
+
+    @cached_property
+    def sizes(self):
+        """Set sizes: core, owned, execute halo, total."""
+        return (self.core_size, self.size, self.exec_size, self.total_size)
+
+    @cached_property
+    def name(self):
+        """User-defined label"""
+        return "GLobalSet"
```
miklos1: Capitalisation.

pyop2/base.py
```diff
@@ -917,7 +984,7 @@ def _process_args(cls, sets, **kwargs):
         try:
             sets = as_tuple(sets, ExtrudedSet)
         except TypeError:
-            sets = as_tuple(sets, Set)
+            sets = as_tuple(sets, (Set, type(None)))
```
miklos1: `type(None)` is usually called `NoneType`, but you need `from types import NoneType` to write that.

pyop2/base.py
```diff
@@ -998,6 +1065,12 @@ def __str__(self):
     def __repr__(self):
         return "MixedSet(%r)" % (self._sets,)
 
+    def __eq__(self, other):
+        try:
+            return self._sets == other._sets
+        except AttributeError:
```
miklos1: Couldn't you write `return type(self) == type(other) and self._sets == other._sets` instead of exception handling?

pyop2/base.py
```diff
@@ -2980,6 +3143,8 @@ class DecoratedMap(Map, ObjectCached):
 
     def __new__(cls, map, iteration_region=None, implicit_bcs=None,
                 vector_index=None):
+        if map is None:
+            return None
```
miklos1: This looks a bit like a hack. Is it used at many places?

pyop2/base.py
```diff
@@ -3205,27 +3371,46 @@ def __init__(self, dsets, maps, name=None, nest=None, block_sparse=None):
         self._rmaps, self._cmaps = zip(*maps)
         self._dsets = dsets
 
-        self.lcomm = self._rmaps[0].comm
-        self.rcomm = self._cmaps[0].comm
-        if self.lcomm != self.rcomm:
-            raise ValueError("Haven't thought hard enough about different left and right communicators")
-        self.comm = self.lcomm
+        if isinstance(dsets[0], GlobalDataSet) or isinstance(dsets[1], GlobalDataSet):
+            # This will cause a trivial memory accounting error (although not a leak).
```
miklos1: Can you elaborate?

pyop2/base.py
dham: So I could write 2 lines instead of 1? Great...

pyop2/base.py
dham: You could, but I'm not convinced that that's actually cleaner.

pyop2/base.py
dham: Well if your map is None (because you are a Global), any corresponding DecoratedMap will also be None.

pyop2/base.py
dham: I'm not sure that's true any more.

pyop2/base.py
miklos1: I think it is better be cautious with exception handling since one might catch exceptions coming from unexpected places. For example, if `self` or `other` have a `_sets` property whose implementation is bugged, or calls some bugged code, e.g. typo in the attribute name of a completely unrelated object, then that exception will be caught here, causing bugs that are very confusing to debug.

PR #504, opened 2016-10-07
Ensure Cython generated files are cleaned correctly

dham:


miklos1:
The change itself looks fine, but who calls it?

dham:
Firedrake-install calls setup.py clean on all our packages. That catches this.

PR #506, opened 2016-11-02
Reuseable parloops

wence-:
Allows ParLoop objects to be reused to repeat the computation.  Will be used in Firedrake to get better strong-scaling in residual and jacobian assembly.  Also, since that side of the change removes assembly caching, we drop dat versioning here (since it is no longer used anywhere).


pyop2/base.py
```diff
@@ -61,8 +59,20 @@
...
 
 
+@contextmanager
+def collecting_loops(val):
+    try:
+        old = LazyComputation.collecting_loops
+        LazyComputation.collecting_loops = val
+        yield
+    finally:
+        LazyComputation.collecting_loops = old
```
miklos1: Dynamic scoping...

wence-: Yeah. Can you think of a better way?

miklos1: Not really, just noting...

Issue #418, opened 2014-12-16
Use absolute imports


wence-:
Lots of the recent coffee migration bugs would have been avoided had we been using absolute imports.  This is recommended in py2 and required in py3k anyway.

wence-:
Done in #507.

Issue #399, opened 2014-09-07
Configuration option for eager host <-> device data transfer


kynan:
Add a configuration option to trigger data transfer eagerly instead of marking data as out of date. This may be advantageous in scenarios where the user knows data is going to be accessed by starting the transfer earlier and potentially overlapping with computation on the host.

wence-:
No longer relevant after #507.

Issue #388, opened 2014-07-17
OpenCL: kernels are not launched asynchronously


kynan:
Currently, OpenCL kernels are enqueued and then we wait for them to finish, which effectively makes them run synchronously.

wence-:
No longer relevant after #507.

Issue #387, opened 2014-07-15
CUDA backend: missing features


kynan:
- [ ] mixed function spaces
- [ ] multi-GPU linear solvers
- [ ] multi-GPU SpMV (required for matrix-free)

wence-:
No longer relevant after #507.

PR #507, opened 2016-11-04
Purge backends

### LINKED ISSUE #414: Expected MacOS build fails confuse new users
Anna thought that PyOP2 was broken because the OpenMP tests fail. It seems likely that other new users would think the same. We should find a way to ensure that "make test" appears to pass on MacOS if all the tests which are expected to work actually do.

wence-:
This is opening for comment and potential merge.

## Summary
- Remove bitrotting, unmaintained, installation routes (install.sh, vagrant file)
- Remove all code generation backends except for the sequential CPU backend
- Remove the Solver object (no-one uses it)

## Rationale

PyOP2 contains a *lot* of technical debt.  This is an attempt to reduce some of it.  I think we have now reached reasonable consensus that the existing implementations for GPUs are *not* the right place to start for high-performance assembly.  Moreover, the OpenMP backend is actively terrible.  If we really want shared memory OpenMP, then going via loop tiling and SLOPE is much better, says @FabioLuporini.

It is likely that projects in the not too distant future will want to touch the code generation of PyOP2 for various reasons: perhaps shared-memory through MPI-3; long vector on KNL; extensions of the extrusion implementation; composable maps.  At the moment, this is a big prospect, not least because one has to ostensibly keep support for (at least some of) the backends.  This hasn't really happened anyway: the OpenCL and cuda backends only really support what was published in Markall et al. (2013), and that in a half-baked fashion.

This change doesn't really address the unmaintainability of code generation, but at least steps towards having less code to change.  Hopefully, it also removes the temptation to start any new work on shared memory from the wrong place.

## Other benefits

- Fixes #414, #418.
- "Fixes" #32, #163, #222, #232, #290, #308, #382, #387, #388, #399


FabioLuporini:
I have no objections against this. In fact, I believe this is a good move.

miklos1:
I tend to think that loo.py is better for generating code for accelerators than the code pasting we now remove.

> This change doesn't really address the unmaintainability of code generation

I have started and almost finished some work on it, but then abandoned it after #493, and now I'm too busy with other stuff, but I plan to pick it up again once the other stuff are done.

miklos1:
Don't you need this patch for Firedrake?
```diff
+++ firedrake/tsfc_interface.py
--- firedrake/tsfc_interface.py
@@ -91,7 +91,7 @@ class TSFCKernel(Cached):
         # FIXME Making the COFFEE parameters part of the cache key causes
         # unnecessary repeated calls to TSFC when actually only the kernel code
         # needs to be regenerated
-        return md5(form.signature() + name + Kernel._backend.__name__
+        return md5(form.signature() + name
                    + str(default_parameters["coffee"])
                    + str(parameters)
                    + str(number_map)).hexdigest(), form.ufl_domains()[0].comm
+++ firedrake/vector.py
--- firedrake/vector.py
@@ -1,7 +1,7 @@
 from __future__ import absolute_import
 import numpy as np
 
-from pyop2 import op2
+from pyop2 import base
 
 from firedrake.petsc import PETSc
 from firedrake.matrix import MatrixBase
@@ -51,7 +51,7 @@ class Vector(object):
         """
         if isinstance(x, Vector):
             self.dat = type(x.dat)(x.dat)
-        elif isinstance(x, op2.base.Dat):  # ugh
+        elif isinstance(x, base.Dat):  # ugh
             self.dat = x
         else:
             raise RuntimeError("Don't know how to build a Vector from a %r" % type(x))

```

wence-:
Forgot to push (now opened).

PR #515, opened 2016-12-06
coffee: FindInstances ---> Find

FabioLuporini:
Depends on coneoproject/COFFEE#109

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -57,7 +57,7 @@ from pyop2.sparsity import build_sparsity
 from pyop2.version import __version__ as version
 
 from coffee.base import Node, FlatBlock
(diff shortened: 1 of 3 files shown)


pyop2/base.py
```diff
@@ -3865,7 +3865,7 @@ def __init__(self, code, name, opts={}, include_dirs=[], headers=[],
         else:
             self._ast = code
             self._code = self._ast_to_c(self._ast, opts)
-            search = FindInstances(ast.FunDecl, ast.FlatBlock).visit(self._ast)
+            search = Find((ast.FunDecl, ast.FlatBlock)).visit(self._ast)
```
wence-: This was previously a bug, yes?

PR #516, opened 2016-12-07
Use kernel name for timing C kernel execution

mlange05:
Profiling information for C kernels of all parloop types (cell,
interior facet, exterior facet) is currently agglomerated into a single
profiling tag. This change enables kernel-specific profiling by using
the kernel name itself as the tag, from which the kernel type can be inferred.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -4186,6 +4186,7 @@ class ParLoop(LazyComputation):
             arg.halo_exchange_end(update_inc=self._only_local)
 
     @collective
+    @timed_function("ParLoopRHaloBegin")
     def reverse_halo_exchange_begin(self):
         """Start reverse halo exchanges (to gather remote data)"""
         if self.is_direct:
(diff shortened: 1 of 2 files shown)


wence-:
This will screw up the formatting in `-log_view` (because the names are long).  What do you want this granularity for?

mlange05:
I'm using it to differentiate between the cell and facet loops triggered within a single expression (the expression is already tagged using petsc events). Maybe a simple map from the long kernel names to something like `ParloopCKernelCell` or `ParloopCKernelInteriorFacet` would help?

wence-:
Use rather the `name` property of the iteration set, I think.

wence-:
Those names are generated in Firedrake as `$MESH_NAME_cells`, `$MESH_NAME_exterior_facets`, `$MESH_NAME_interior_facets`, where `$MESH_NAME` is almost always `None`.  You could also clean up on that side an make them by `Cells`, `IntFacets`, `ExtFacets`.  I think that would do the trick.

mlange05:
Thanks for the pointers, that indeed looks much nicer. The relevant firedrake PR is firedrakeproject/firedrake#960.

Issue #518, opened 2016-12-13
Node local compilation


nickaj:
If running on a large system where writing to a central filesystem is slow, it makes sense to try to compile per-node (say in /tmp) rather than than one process compiling to the fs then everyone reading from that.

nickaj:
A PR to do this should be forthcoming tomorrow morning.

wence-:
Done in #539.

PR #519, opened 2016-12-14
Add code to perform node-local compilation.

nickaj:
Add code to perform node-local compilation based either on MPI3 shared memory regions
or using a file-based method if MPI3 is not available. This may help to speed compilation
in cases where access to a shared file-system is slow.

This addresses Issue #518

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -31,12 +31,14 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
 # OF THE POSSIBILITY OF SUCH DAMAGE.
 
-from __future__ import absolute_import
+from __future__ import absolute_import, print_function, division
+import six
+from six.moves import input
+
 import os
 import subprocess
 import sys
 import ctypes
-import random
 from hashlib import md5
 
 from pyop2.mpi import MPI, collective, COMM_WORLD
@@ -50,6 +52,7 @@ def _check_hashes(x, y, datatype):
     if x == y:
         return x
     return False
+
 
 _check_op = MPI.Op.Create(_check_hashes, commute=True)
 
@@ -91,12 +94,12 @@ class Compiler(object):
         library."""
 
         # Determine cache key
-        hsh = md5(src)
-        hsh.update(self._cc)
+        hsh = md5(six.b(src))
+        hsh.update(six.b(self._cc))
         if self._ld:
-            hsh.update(self._ld)
-        hsh.update("".join(self._cppargs))
-        hsh.update("".join(self._ldargs))
+            hsh.update(six.b(self._ld))
+        hsh.update(six.b("".join(self._cppargs)))
+        hsh.update(six.b("".join(self._ldargs)))
 
         basename = hsh.hexdigest()
 
@@ -108,59 +111,42 @@ class Compiler(object):
         # Link into temporary file, then rename to shared library
         # atomically (avoiding races).
         tmpname = os.path.join(cachedir, "%s_p%d.so.tmp" % (basename, pid))
-        singleton = False
-
-        # To aid in compilation speed, compile only on one MPI rank per node.
-        if MPI.VERSION >= 3:
-            # If MPI3 library is available, split the communicator based on shared memory regions.
-            newcomm = self.comm.Split_type(MPI.COMM_TYPE_SHARED)
-            debug("Using MPI3 on rank ", newcomm.rank)
-            if newcomm.rank==0:
-                singleton = True
-        else:
-            # Otherwise, copy the communicator and use file-based methods to isolate one rank per node.
-            newcomm = self.comm
-            debug("Not using MPI3 on %d" % newcomm.rank)
-            singleton = self.single_out_a_rank(newcomm)
-            MPI.Comm.Barrier(newcomm)
-            
 
         if configuration['check_src_hashes'] or configuration['debug']:
-            matching = newcomm.allreduce(basename, op=_check_op)
+            matching = self.comm.allreduce(basename, op=_check_op)
             if matching != basename:
                 # Dump all src code to disk for debugging
                 output = os.path.join(cachedir, "mismatching-kernels")
-                srcfile = os.path.join(output, "src-rank%d.c" % newcomm.rank)
-                if newcomm.rank == 0:
+                srcfile = os.path.join(output, "src-rank%d.c" % self.comm.rank)
+                if self.comm.rank == 0:
                     if not os.path.exists(output):
                         os.makedirs(output)
-                newcomm.barrier()
+                self.comm.barrier()
                 with open(srcfile, "w") as f:
                     f.write(src)
-                newcomm.barrier()
+                self.comm.barrier()
                 raise CompilationError("Generated code differs across ranks (see output in %s)" % output)
-            
         try:
             # Are we in the cache?
-            return ctypes.CDLL(soname)    
+            return ctypes.CDLL(soname)
         except OSError:
             # No, let's go ahead and build
-            if singleton:
+            if self.comm.rank == 0:
                 # No need to do this on all ranks
                 if not os.path.exists(cachedir):
                     os.makedirs(cachedir)
                 logfile = os.path.join(cachedir, "%s_p%d.log" % (basename, pid))
                 errfile = os.path.join(cachedir, "%s_p%d.err" % (basename, pid))
                 with progress(INFO, 'Compiling wrapper'):
-                    with file(cname, "w") as f:
+                    with open(cname, "w") as f:
                         f.write(src)
                     # Compiler also links
                     if self._ld is None:
                         cc = [self._cc] + self._cppargs + \
                              ['-o', tmpname, cname] + self._ldargs
                         debug('Compilation command: %s', ' '.join(cc))
-                        with file(logfile, "w") as log:
-                            with file(errfile, "w") as err:
+                        with open(logfile, "w") as log:
+                            with open(errfile, "w") as err:
                                 log.write("Compilation command:\n")
                                 log.write(" ".join(cc))
                                 log.write("\n\n")
@@ -186,8 +172,8 @@ Compile errors in %s""" % (e.cmd, e.returncode, logfile, errfile))
                         ld = self._ld.split() + ['-o', tmpname, oname] + self._ldargs
                         debug('Compilation command: %s', ' '.join(cc))
                         debug('Link command: %s', ' '.join(ld))
-                        with file(logfile, "w") as log:
-                            with file(errfile, "w") as err:
+                        with open(logfile, "w") as log:
+                            with open(errfile, "w") as err:
                                 log.write("Compilation command:\n")
                                 log.write(" ".join(cc))
                                 log.write("\n\n")
@@ -220,33 +206,9 @@ Compile errors in %s""" % (e.cmd, e.returncode, logfile, errfile))
                     # Atomically ensure soname exists
                     os.rename(tmpname, soname)
             # Wait for compilation to complete
-            newcomm.barrier()
             self.comm.barrier()
             # Load resulting library
             return ctypes.CDLL(soname)
-
-    @collective
-    def single_out_a_rank(self, comm):
-        rank = MPI.Comm.Get_rank(comm)
-        debug('Using file-based mechanism on %d', rank)
-        fname =  os.path.join("/tmp", "fb.%d" % rank)
-
-        with open(fname, 'w+') as f:
-            f.write("This is a file.")
-
-        MPI.Comm.Barrier(comm)
-        tbb = []
-        for dl,sl,fl in os.walk("/tmp"):
-            for fi in fl:
-                if str(fi).split(".")[0] == "fb":
-                    tbb.append(str(fi).split(".")[1])
-
-        MPI.Comm.Barrier(comm)
-        os.remove(fname)
-        if str(min(tbb)) == str(rank):
-            return True
-        else:
-            return False
 
 
 class MacCompiler(Compiler):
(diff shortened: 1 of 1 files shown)


pyop2/compilation.py
```diff
@@ -31,14 +31,12 @@
 # ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED
 # OF THE POSSIBILITY OF SUCH DAMAGE.
 
-from __future__ import absolute_import, print_function, division
-import six
-from six.moves import input
-
+from __future__ import absolute_import
```
wence-: I think you back-merged something the wrong way?  Can you rebase your changes on current master (so that we're not destroying py3 compatibility)

pyop2/compilation.py
```diff
@@ -111,42 +108,59 @@ def get_so(self, src, extension):
...
         # atomically (avoiding races).
         tmpname = os.path.join(cachedir, "%s_p%d.so.tmp" % (basename, pid))
+        singleton = False
+
+        # To aid in compilation speed, compile only on one MPI rank per node.
+        if MPI.VERSION >= 3:
+            # If MPI3 library is available, split the communicator based on shared memory regions.
+            newcomm = self.comm.Split_type(MPI.COMM_TYPE_SHARED)
+            debug("Using MPI3 on rank ", newcomm.rank)
+            if newcomm.rank==0:
```
wence-: Spaces around "==" (see errors from `make lint`).

pyop2/compilation.py
```diff
@@ -206,10 +220,34 @@ def get_so(self, src, extension):
...
+            newcomm.barrier()
             self.comm.barrier()
             # Load resulting library
             return ctypes.CDLL(soname)
 
+    @collective
+    def single_out_a_rank(self, comm):
+        rank = MPI.Comm.Get_rank(comm)
+        debug('Using file-based mechanism on %d', rank)
+        fname =  os.path.join("/tmp", "fb.%d" % rank)
```
wence-: Previously, we used the configuration variable `"cache_dir"`, can you use that here?  Please also use `mkstemp` or whatever it's called to generate the prefix to the name (then broadcast it).

pyop2/compilation.py
```diff
@@ -206,10 +220,34 @@ def get_so(self, src, extension):
...
+            f.write("This is a file.")
+
+        MPI.Comm.Barrier(comm)
+        tbb = []
+        for dl,sl,fl in os.walk("/tmp"):
+            for fi in fl:
+                if str(fi).split(".")[0] == "fb":
+                    tbb.append(str(fi).split(".")[1])
+
+        MPI.Comm.Barrier(comm)
```
wence-: I think this barrier is unnecessary: you're not going to try and communicate out of band any more.

pyop2/compilation.py
```diff
@@ -206,10 +220,34 @@ def get_so(self, src, extension):
                     # Atomically ensure soname exists
                     os.rename(tmpname, soname)
             # Wait for compilation to complete
+            newcomm.barrier()
             self.comm.barrier()
             # Load resulting library
             return ctypes.CDLL(soname)
 
+    @collective
+    def single_out_a_rank(self, comm):
```
wence-: This doesn't need `self` at all, just make it a free function.

pyop2/compilation.py
```diff
@@ -111,42 +108,59 @@ def get_so(self, src, extension):
...
+            newcomm = self.comm.Split_type(MPI.COMM_TYPE_SHARED)
+            debug("Using MPI3 on rank ", newcomm.rank)
+            if newcomm.rank==0:
+                singleton = True
+        else:
+            # Otherwise, copy the communicator and use file-based methods to isolate one rank per node.
+            newcomm = self.comm
+            debug("Not using MPI3 on %d" % newcomm.rank)
+            singleton = self.single_out_a_rank(newcomm)
+            MPI.Comm.Barrier(newcomm)
```
wence-: Why this barrier?

wence-:
Currently you do this determination once per compilation.  It would be better to do this once per communicator.  What I recommend (you can learn about MPI attributes!) is that you inspect the incoming communicator, find the internal PyOP2 communicator on it (see pyop2/mpi.py for how this works).  Then, you inspect *that* communicator to see if it has an appropriately split communicator attached to it.  If yes, just use that.  If no, split it and set the attribute.  So then the process is something like:

```
compile(src, comm=comm)
  compile_comm = get_compile_comm(comm)
  compile_on(src, compile_comm)
```

You then only do the "expensive" determination once per user communicator, not once per call.

pyop2/compilation.py
nickaj: I did the fork yesterday after our call so I'm not sure what happened there. I'll try a rebase to see what gives.

pyop2/compilation.py
```diff
@@ -111,42 +108,59 @@ def get_so(self, src, extension):
...
         tmpname = os.path.join(cachedir, "%s_p%d.so.tmp" % (basename, pid))
+        singleton = False
+
+        # To aid in compilation speed, compile only on one MPI rank per node.
+        if MPI.VERSION >= 3:
+            # If MPI3 library is available, split the communicator based on shared memory regions.
+            newcomm = self.comm.Split_type(MPI.COMM_TYPE_SHARED)
+            debug("Using MPI3 on rank ", newcomm.rank)
+            if newcomm.rank==0:
+                singleton = True
```
nickaj: I could be persuaded either way. If we restrict same disk to be same node-local temp file-system, then I don't think it's an unreasonable assumption.

pyop2/compilation.py
nickaj: Sure.

pyop2/compilation.py
wence-: What I mean is, can you imagine a situation where this would fail in hardware you've seen?

pyop2/compilation.py
```diff
@@ -54,6 +55,31 @@ def _check_hashes(x, y, datatype):
     return False
 
 
+@collective
+def single_out_a_rank(comm):
+    rank = MPI.Comm.Get_rank(comm)
+    debug('Using file-based mechanism on %d', rank)
+    fname = os.path.join("/tmp", "fb.%d" % rank)
+
+    with open(fname, 'w+') as f:
```
miklos1: You probably don't need the `+`, do you?

pyop2/compilation.py
```diff
@@ -54,6 +55,31 @@ def _check_hashes(x, y, datatype):
...
+        f.write("This is a file.")
+
+    MPI.Comm.Barrier(comm)
+    tbb = []
+    # for dl, sl, fl in os.walk("/tmp"):
+    #     for fi in fl:
```
miklos1: Please remove commented out code!

pyop2/compilation.py
```diff
@@ -54,6 +55,31 @@ def _check_hashes(x, y, datatype):
...
+    files = glob.glob("/tmp/fb.[0-9]+")
+    for fi in files:
+        if str(fi).split(".")[0] == "fb":
+            tbb.append(str(fi).split(".")[1])
+
+    os.remove(fname)
+    if str(min(tbb)) == str(rank):
```
miklos1: It seems to be that `tbb` is a list of strings rather than a list of integers, so this will lexicographically smallest number rather than the smallest number. Though this shouldn't cause a problem.

pyop2/compilation.py
```diff
@@ -54,6 +55,31 @@ def _check_hashes(x, y, datatype):
...
+        return True
+    else:
+        return False
```
miklos1:
```python
if condition:
    return True
else:
    return False
```
is unnecessarily verbose, please write instead:
```python
return condition
```

pyop2/compilation.py
```diff
@@ -112,26 +138,38 @@ def get_so(self, src, extension):
...
         if configuration['check_src_hashes'] or configuration['debug']:
-            matching = self.comm.allreduce(basename, op=_check_op)
+            matching = newcomm.allreduce(basename, op=_check_op)
             if matching != basename:
                 # Dump all src code to disk for debugging
                 output = os.path.join(cachedir, "mismatching-kernels")
-                srcfile = os.path.join(output, "src-rank%d.c" % self.comm.rank)
-                if self.comm.rank == 0:
+                srcfile = os.path.join(output, "src-rank%d.c" % newcomm.rank)
+                if newcomm.rank == 0:
```
miklos1: Assuming rank zero is probably not the right thing to do for the non-MPI3 case.

pyop2/compilation.py
```diff
@@ -112,26 +138,38 @@ def get_so(self, src, extension):
...
+                newcomm.barrier()
                 raise CompilationError("Generated code differs across ranks (see output in %s)" % output)
+
         try:
             # Are we in the cache?
             return ctypes.CDLL(soname)
         except OSError:
             # No, let's go ahead and build
-            if self.comm.rank == 0:
+            if singleton:
```
miklos1: Yes, this one I agree with.

pyop2/compilation.py
```diff
@@ -206,6 +244,7 @@ def get_so(self, src, extension):
                     # Atomically ensure soname exists
                     os.rename(tmpname, soname)
             # Wait for compilation to complete
+            newcomm.barrier()
             self.comm.barrier()
```
miklos1: Do we really need a double barrier here?

pyop2/compilation.py
```diff
@@ -54,6 +55,31 @@ def _check_hashes(x, y, datatype):
     return False
 
 
+@collective
+def single_out_a_rank(comm):
+    rank = MPI.Comm.Get_rank(comm)
+    debug('Using file-based mechanism on %d', rank)
+    fname = os.path.join("/tmp", "fb.%d" % rank)
```
wence-: Use `mktmp` or whatever it's called to get a guaranteed unique directory name, broadcast that, and then write into the subdirectory.  Otherwise, three things:

1. You might have stuff left around from a previous run
2. /tmp might not be the right place
3. Someone else might be writing to `/tmp/fb.rank`

pyop2/compilation.py
```diff
@@ -54,6 +55,31 @@ def _check_hashes(x, y, datatype):
     return False
 
 
+@collective
+def single_out_a_rank(comm):
+    rank = MPI.Comm.Get_rank(comm)
```
wence-: AKA `comm.rank`

pyop2/compilation.py
```diff
@@ -54,6 +55,31 @@ def _check_hashes(x, y, datatype):
...
+    debug('Using file-based mechanism on %d', rank)
+    fname = os.path.join("/tmp", "fb.%d" % rank)
+
+    with open(fname, 'w+') as f:
+        f.write("This is a file.")
+
+    MPI.Comm.Barrier(comm)
```
wence-: AKA `comm.barrier()`.

pyop2/compilation.py
```diff
@@ -54,6 +55,31 @@ def _check_hashes(x, y, datatype):
...
+    tbb = []
+    # for dl, sl, fl in os.walk("/tmp"):
+    #     for fi in fl:
+    files = glob.glob("/tmp/fb.[0-9]+")
+    for fi in files:
+        if str(fi).split(".")[0] == "fb":
+            tbb.append(str(fi).split(".")[1])
+
+    os.remove(fname)
```
wence-: I think you need a barrier before removing the file.  Otherwise rank A may remove its file before rank B reads it.  (It's like shared memory consistency models, only worse, because the hardware gives you even fewer guarantees...).

pyop2/compilation.py
```diff
@@ -112,26 +138,38 @@ def get_so(self, src, extension):
...
+        if MPI.version >= 3:
+            # Assume that the library supports flashy MPI3 features
+            newcomm = self.comm.Split_type(MPI.COMM_TYPE_SHARED)
+            debug("Using MPI3 on rank ", newcomm.rank)
+            if newcomm.rank == 0:
+                singleton = True
+        else:
+            newcomm = self.comm
+            debug("Not using MPI3 on %d" % newcomm.rank)
+            singleton = self.single_out_a_rank(newcomm)
```
wence-: I would prefer if the logic was:

```
newcomm = split_comm(comm)
```

where `split_comm` either splits based on `MPI_COMM_TYPE_SHARED`, or else on the file-based system.

wence-: `single_out_a_rank` is a free function, right?  So no `self` here.

pyop2/compilation.py
wence-: I would hope not.

pyop2/compilation.py
```diff
@@ -112,26 +138,38 @@ def get_so(self, src, extension):
...
+
         if configuration['check_src_hashes'] or configuration['debug']:
-            matching = self.comm.allreduce(basename, op=_check_op)
+            matching = newcomm.allreduce(basename, op=_check_op)
```
wence-: In the non-MPI three case, these barriers are all bigger than they need to be (because the comm is COMM_WORLD).  Hence my suggestion to split in both cases.

pyop2/compilation.py
```diff
@@ -112,26 +138,38 @@ def get_so(self, src, extension):
         # atomically (avoiding races).
         tmpname = os.path.join(cachedir, "%s_p%d.so.tmp" % (basename, pid))
 
+        if MPI.version >= 3:
+            # Assume that the library supports flashy MPI3 features
+            newcomm = self.comm.Split_type(MPI.COMM_TYPE_SHARED)
```
wence-: Unfortunately, mpi4py doesn't garbage collect much (in particular communicators).  So this leaks comm handles (one per compile call).  So you either need to call newcomm.Free() later, or else (if we don't want to do all the determination every time we compile things) manage the split communicators yourself using MPI attributes.

PyOP2 already has code for managing "internal" communicators (when the user passes in a communicator, we check to see if we've already dupped it, if not, we dup it and stash the dupped communicator using attributes.  See that code in `pyop2/mpi.py`.  It should be readily adaptable for your usecase (you just need to change the COMM_DUP call to COMM_SPLIT).

nickaj:
Closing to allow me to push work & branches around.

miklos1:
> push work & branches around

I'm not sure what you mean, nor do I understand why you need to close this pull request to do that.

PR #520, opened 2017-01-30
Allow passing layer argument to kernel

thomasgibson:
This PR does exactly what the title says: it allows for the layer argument (during an extrusion iteration) to be passed to the kernel in the wrapper.

The parameter is defaulted to `False` and should not affect any greater-Firedrake work unless the keyword argument `pass_layer_arg=True` is passed to the parloop. Slate will require this kwarg when handling extruded facet integrals.

### DIFF
--- a/test/unit/test_extrusion.py
+++ b/test/unit/test_extrusion.py
@@ -377,6 +377,25 @@ void comp_vol(double A[1], double *x[], double *y[])
                      dat.dataset.set, dat(op2.INC))
         assert numpy.allclose(dat.data[:], 1.0)
 
+    def test_extruded_layer_arg(self, elements, dat_coords,
+                                dat_field, coords_map, field_map,
+                                dat_f):
+        """Tests that the layer argument is being passed when prompted
+        to in the parloop."""
+
+        kernel_blah = """void kernel_blah(double* x[], int layer_arg){
+                                                 x[0][0] = layer_arg;
+                                              }\n"""
+
+        op2.par_loop(op2.Kernel(kernel_blah, "kernel_blah"),
+                     elements, dat_f(op2.WRITE, field_map),
+                     pass_layer_arg=True)
+        end = layers - 1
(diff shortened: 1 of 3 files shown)


wence-:
Is it possible to add a test of this functionality in the pyop2 test suite?  `test/unit/...extrusion.py` ?

thomasgibson:
I can try to cook something up. Should I just write up a random `op2.Kernel` that does some arithmetic on a dat and the layer index on an extruded set?

wence-:
Basically yes.

wence-:
For example, you could write a kernel that writes the layer into a dat, and then check that you get the right answer.

test/unit/test_extrusion.py
```diff
@@ -377,6 +377,25 @@ def test_direct_loop_inc(self, xtr_nodes):
...
+        kernel_blah = """void kernel_blah(double* x[], int layer_arg){
+                                                 x[0][0] = layer_arg;
+                                              }\n"""
+
+        op2.par_loop(op2.Kernel(kernel_blah, "kernel_blah"),
+                     elements, dat_f(op2.WRITE, field_map),
+                     pass_layer_arg=True)
+        end = layers - 1
+        start = 0
+        ref = np.array(range(start, end))
```
miklos1: AKA `np.arange`

test/unit/test_extrusion.py
```diff
@@ -377,6 +377,25 @@ def test_direct_loop_inc(self, xtr_nodes):
...
+        assert [dat_f.data[10*n:10*(n+1)] == ref
```
miklos1: What is 10?

thomasgibson: number of extruded mesh layers (instances of the base mesh)

thomasgibson:
Test added/edited. Let me know if it's alright and ready to merge.

PR #521, opened 2017-02-09
Support 64 bit ints as map values

wence-:
Since PETSc only has one int type, we must use that everywhere (not just for the only place it's really necessary, which is global numbering).

Do so by determining the int type from petsc4py and using that everywhere.

### DIFF
--- a/pyop2/sequential.py
+++ b/pyop2/sequential.py
@@ -36,11 +36,11 @@ from __future__ import absolute_import, print_function, division
 from six.moves import range, zip
 
 import os
-import ctypes
 from textwrap import dedent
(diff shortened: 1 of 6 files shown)


wence-:
Do not merge until matching firedrake side is ready (this breaks some cases).

pyop2/sequential.py
```diff
@@ -657,6 +673,7 @@ def compile(self):
         #include <petsc.h>
         #include <stdbool.h>
         #include <math.h>
+        #include <inttypes.h>
```
miklos1: `stdint.h` should be sufficient, no?

PR #523, opened 2017-03-17
compilation: automate applying compiler bug workarounds

wence-:
Rather than hard-coding for everyone, select correct "bugfix" flags
depending on the compiler version.

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -39,7 +39,10 @@ import os
 import subprocess
 import sys
 import ctypes
+import collections
 from hashlib import md5
+from distutils import version
+
 
 from pyop2.mpi import MPI, collective, COMM_WORLD
 from pyop2.configuration import configuration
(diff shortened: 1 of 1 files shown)


miklos1:
More "workaround" than "bugfix".

pyop2/compilation.py
```diff
@@ -79,10 +117,31 @@ def __init__(self, cc, ld=None, cppargs=[], ldargs=[],
...
+    def bugfix_cflags(self):
+        """Flags to work around bugs in compilers."""
+        compiler, ver = self.compiler_version
+        if compiler == "gcc":
+            if version.StrictVersion("4.8.0") <= ver < version.StrictVersion("4.9.0"):
+                # GCC bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=61068
+                return ["-fno-ivopts"]
+            if version.StrictVersion("6.0.0") <= ver < version.StrictVersion("7.0.1"):
+                # GCC bug https://gcc.gnu.org/bugzilla/show_bug.cgi?id=79920
+                return ["-fno-tree-loop-vectorize"]
```
dorugeber: `-fno-tree-loop-vectorize` or `-fno-tree-vectorize`?

wence-: The former appeared to work for me?

dorugeber: Okay, the GCC guy on the bug report mentioned the latter, that's all.

PR #526, opened 2017-06-08
Owner computes

wence-:
Move to an owner-computes model, rather than a redundant computation model.

The rationale for this is that correct redundant computation in the presence of facet integrals requires depth-2 halos if there is only partial discontinuity.

This passes the PyOP2 test suite, but is not ready for merge yet, because we need some PETSc changes, as well as Firedrake ones.  But comments welcome.

In particular, I have cargo-culted the loop fusion changes @tj-sun, @FabioLuporini, do they look plausible?


FabioLuporini:
They look plausible.

Unfortunately, w/o redundant computation, this is gonna destroy sparse tiling (yes, I'm still occasionally on that, I haven't given up yet -- in fact, I'm closer and closer to the results I needed). But yes, I see the point for these changes, I can't say I have crazy speed-ups from sparse tiling, so this looks reasonable to me. 

Incidentally, I was talking with @mlange05 yesterday about resurrecting the deep halo work in Firedrake. We have had deep halos for a while, but they never made master. Deep halos allow sparse tiling over longer sequences of loops

wence-:
My feeling with sparse tiling is that we should figure out how to do it by doing the unstructured equivalent of diamond tiling: no redundant compute anywhere, just more complicated.

FabioLuporini:
As far as I remember/understood it, diamond tiling would need redundant computation just like sparse tiling (unless introducing explicit synchros/comms between consecutive loops -- an approach which would also be possible with sparse tiling, but that looked to me as crazy complicated). 

But no worries: my concern was merely personal

FabioLuporini:
And I have got this (not so mild, not so strong) hypothesis: sparse tiling sort of is the unstructured equivalent of diamond tiling

wence-:
No, what you do is shrink the owned region, then you exchange along a space-time halo, and fill in the gaps. So there's no redundant compute.

FabioLuporini:
You can do that with sparse tiling too. But then don't you have to pay N extra synchronising communications, one for each tiled loop (or timestep, if you have only one loop in your timestep), with marginal computation/communication overlap? 

If you look at the figures in my thesis, you see that the owned regions do grow inside; however, the presence of thick layers allows the avoidance of these extra synchronisations

FabioLuporini:
IOW: sparse tiling is a general way of blocking a driving the growing/shrinking process; that can be composed with different distributed-memory parallelism schemes. In my implementation, only one such scheme was supported: the one with thick layers

wence-:
So if you only have "space" halos yes. But imagine computing on a processor domain. Now I can do the next loop on a smaller region  but on the second step, I haven't computed all of the dofs. So now I do a halo exchange along the diagonal slice, go back to the first step, and fill in the gaps with an expanding, rather than contracting, tile. E.g. See the figures in https://arxiv.org/abs/1410.3060

FabioLuporini:
Could you contextualise this explanation with Figure 4.2 of the paper you linked?

From the paper:
"Regular diamond tiles are used at the boundary of subdomains, with the difference of performing communication before and after the tile update."
The way I see it (but I might be simply overthinking it) is that you have two options:
- Either you send "more data" so that a boundary tile can compute the entire slice of missing "DoFs" (this "more data" would be proportionate to the depth of the space/time tiles)
- Or you do the following: 
  1. get halo from neighbouring process;
  2. compute owned "DoFs" for loop_0
  3. send stuff computed at step 2. to the neighbouring process
  4. restart from 1, but now do loop_1 
Is what you're saying one of these two schemes? I don't know how these two schemes fit in that paper sentence that I quoted above. Might be I'm simply missing the third option

I think you can carve diamond tiles with the base infrastructure of sparse tiling, by choosing a suitable "seed loop" and "coloring scheme" (see draft paper I sent you and here: https://pdfs.semanticscholar.org/b55d/12f510539fa78b0bddbef062e46b4f7b2e58.pdf); however, my feeling is that obtaining these diamond tiles is orthogonal to how you choose to deal with MPI

The last thought I have is about the necessity of fancy tiling schemes at all. If you only rely on MPI for parallelism (note: just pure MPI; forget about mixed-parallelism solutions such as MPI+OpenMP or MPI+MPI) then all tiles but the boundary ones of a given process X can be executed atomically and sequentially, without spilling to memory when "jumping" from one tile to the next one (the problem that diamond tiling tries to "fix").

Apologies if this went too far from the objective of the PR; in which case, we can continue this discussion offline

wence-:
So it is indeed neither of those proposed schemes.  On loop_0, you have up to date ghost regions, so you compute the full tile.  The loop_1 doesn't have up to date ghosts, so you compute a smaller tile.  Now loop_2 computes a still smaller tile.  So you get a kind of pyramid of correctly computed values in space-time.

OK, so now what do you do to fill in the uncomputed values: you do a halo exchange along the diagonal space-time boundary.  Then you go back in time to loop_0, compute the new data you need using the halos, then do loop_1, propagating the owned data into the uncomputed region, then loop_2 and so on.  So you fill in the gap with an upside down space-time triangle.  And repeat.

FabioLuporini:
This was discussed offline. A long story short: doable but very complicated

pyop2/base.py
```diff
@@ -2723,13 +2567,25 @@ def zero(self):
...
+        pass
+
+    @collective
+    def local_to_global_begin(self):
+        """Dummy halo operation for the case in which a :class:`Global` forms
+        part of a :class:`MixedDat`."""
+        pass
+
+    @collective
+    def local_to_global_end(self):
```
wence-: Need to add `insert_mode` argument I think.

pyop2/sparsity.pyx
```diff
@@ -56,304 +54,119 @@ cdef extern from "petsc.h":
...
+                mat = preallocator.getLocalSubMatrix(isrow=rset.local_ises[i],
+                                                     iscol=cset.local_ises[j])
+                fill_with_zeros(mat, (r.cdim, c.cdim),
+                                maps,
+                                set_diag=((i == j) and sparsity._has_diagonal))
+                mat.assemble()
+                preallocator.restoreLocalSubMatrix(isrow=rset.local_ises[i],
+                                                   iscol=cset.local_ises[j],
+                                                   submat=mat)
+        preallocator.assemble()
```
wence-: We use new PETSc technology to build preallocation, which greatly simplifies things.

pyop2/base.py
```diff
@@ -1363,154 +1283,59 @@ def __repr__(self):
...
 
-        for source, receives in self.receives.items():
-            assert (receives >= s.size).all() and \
-                (receives < s.total_size).all(), \
-                "Halo receive from %d is invalid (not in halo elements)" % \
-                source
+        :arg dat: The :class:`Dat` to exchange.
+        :arg insert_mode: The insertion mode.
+        """
+        pass
```
wence-: We basically only ever use the Firedrake halo, so make this an abstract class and change the naming convention to more closely match PETSc.

pyop2/base.py
```diff
@@ -2218,30 +2030,52 @@ def __itruediv__(self, other):
...
-        """Begin halo exchange.
+    def global_to_local_begin(self, insert_mode):
+        """Begin a halo exchange from global to ghosted representation.
+
+        :kwarg insert_mode:  insertion mode (an access descriptor)"""
+        halo = self.dataset.halo
+        if halo is None:
+            return
+        if insert_mode in [READ, RW] and not self.halo_valid:
+            halo.global_to_local_begin(self, WRITE)
```
wence-: I realise the `insert_mode` name changes meaning from this method to the halo call.  To the halo it means, do we write or sum into the array.  Here, insert_mode is actually access_mode.

FabioLuporini:
BTW, any idea of what's the performance gain from switching to this model at small, medium, and large scale ?

wence-:
No, but it fixes a bug for us. I expect it will help a bit with strong scaling in the extruded cases

setup.py
```diff
@@ -123,7 +123,7 @@ class sdist(_sdist):
     def run(self):
         # Make sure the compiled Cython files in the distribution are up-to-date
         from Cython.Build import cythonize
-        cythonize(sparsity_sources, language="c++", include_path=includes)
+        cythonize(sparsity_sources, language="c", include_path=includes)
```
miklos1: :+1:

pyop2/sparsity.pyx
```diff
@@ -56,304 +54,119 @@ cdef extern from "petsc.h":
                                  PetscScalar*, PetscInsertMode)
     int MatSetValuesLocal(PETSc.PetscMat, PetscInt, PetscInt*, PetscInt, PetscInt*,
                           PetscScalar*, PetscInsertMode)
+    int MatPreallocatorPreallocate(PETSc.PetscMat, PetscBool, PETSc.PetscMat)
+    int MatXAIJSetPreallocation(PETSc.PetscMat, PetscInt, const PetscInt[], const PetscInt[],
+                                const PetscInt[], const PetscInt[])
 
+cdef extern from "petsc/private/matimpl.h":
+    struct _p_Mat:
+        void *data
```
miklos1: Okay...

pyop2/sparsity.pyx
miklos1: :+1:

pyop2/base.py
```diff
@@ -464,35 +465,46 @@ def _uses_itspace(self):
         return self._is_mat or isinstance(self.idx, IterationIndex)
 
     @collective
-    def halo_exchange_begin(self, update_inc=False):
```
miklos1: Okay, so since we are on an `Arg`, we already know what the access descriptor is, so no need for `update_inc=...`.

pyop2/base.py
miklos1: Sounds good.

pyop2/sparsity.pyx
wence-: This is how the PETSc runtime delegation pattern is implemented.  See https://github.com/firedrakeproject/petsc/blob/firedrake/include/petsc/private/matimpl.h#L375

pyop2/sparsity.pyx
wence-: An even better way to do this would be to execute a `par_loop` with dummy kernel of appropriate size, but I'll save that for another time.

PR #527, opened 2017-06-15
Add a vec_wo context manager to Dats

wence-:
Avoids some computation and halo exchanges in the case that we're
immediately going to overwrite all the data.

### DIFF
--- a/pyop2/petsc_base.py
+++ b/pyop2/petsc_base.py
@@ -312,27 +312,25 @@ class MixedDataSet(DataSet, base.MixedDataSet):
 class Dat(base.Dat):
 
     @contextmanager
-    def vec_context(self, readonly=True):
+    def vec_context(self, access):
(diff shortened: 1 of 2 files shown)


pyop2/petsc_base.py
```diff
@@ -412,15 +418,24 @@ def vec(self):
...
     @collective
     def vec_ro(self):
         """Context manager for a PETSc Vec appropriate for this Dat.
 
         You're not allowed to modify the data you get back from this view."""
-        return self.vecscatter()
+        return self.vecscatter(access=base.READ)
 
 
 class Global(base.Global):
```
thomasgibson: Shouldn't `Global` also have a `vec_wo` property?

wence-: Oh yeah, I made that change locally, but did not push, now done.

PR #531, opened 2017-07-31
Define infix operators on Globals

dham:
This fixes issues where operations on Firedrake Functions in the R space assume that the usual operators are defined for Globals.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -2792,8 +2792,6 @@ class Global(DataCarrier, _EmptyDataMixin):
         """Pointwise division or scaling of fields."""
         return self._op(other, operator.truediv)
 
-    __div__ = __truediv__  # Python 2 compatibility
-
     def __iadd__(self, other):
         """Pointwise addition of fields."""
         return self._iop(other, operator.iadd)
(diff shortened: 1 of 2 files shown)


pyop2/base.py
```diff
@@ -2737,6 +2737,84 @@ def halo_exchange_end(self):
...
+
+    def __imul__(self, other):
+        """Pointwise multiplication or scaling of fields."""
+        return self._iop(other, operator.imul)
+
+    def __itruediv__(self, other):
+        """Pointwise division or scaling of fields."""
+        return self._iop(other, operator.itruediv)
+
+    __idiv__ = __itruediv__  # Python 2 compatibility
```
miklos1: Are we still supporting Python 2 in PyOP2?

dham: Well we haven't disabled it elsewhere, so not putting it in at this point would create an undocumented breakage. However I agree, we should drop Python 2.7 support (but do it in one fell swoop).

miklos1: See #530.

dham: Happy now?

test/unit/test_globals.py
```diff
@@ -0,0 +1,46 @@
+# This file is part of PyOP2
+#
+# PyOP2 is Copyright (c) 2012, Imperial College London and
```
miklos1: Wrong date.

pyop2/base.py
miklos1: Yes.

PR #534, opened 2017-09-19
Compilation: wait more to make sure the .so file exists on disk

tkarna:
As mentioned in firedrakeproject/firedrake#1145, compilation in parallel may fail in case some ranks do not see the .so file immediately.

This patch adds a check of the existence of the file, and an additional wait up to 60 s. Should not introduce much overhead in systems where the issue does not occur; there's only one additional `os.path.isfile` check at compile time.

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -39,6 +39,7 @@ import ctypes
 import collections
 from hashlib import md5
 from distutils import version
+from time import sleep
 
 
 from pyop2.mpi import MPI, collective, COMM_WORLD
@@ -266,6 +267,12 @@ Compile errors in %s""" % (e.cmd, e.returncode, logfile, errfile))
                     os.rename(tmpname, soname)
             # Wait for compilation to complete
             self.comm.barrier()
+            if not os.path.isfile(soname):
+                # Wait more if .so file is not visible on disk
+                for i in range(60):
+                    sleep(1.0)
+                    if os.path.isfile(soname):
+                        break
             # Load resulting library
             return ctypes.CDLL(soname)
 


pyop2/compilation.py
```diff
@@ -266,6 +267,12 @@ def get_so(self, src, extension):
                     os.rename(tmpname, soname)
             # Wait for compilation to complete
             self.comm.barrier()
+            if not os.path.isfile(soname):
+                # Wait more if .so file is not visible on disk
+                for i in range(60):
+                    sleep(1.0)
+                    if os.path.isfile(soname):
+                        break
```
miklos1: I wonder if there is portable way that does not involve polling... some timed out waiting for change in the directory or anything similar.

wence-: Doubtful, you might think something like inotify would work, but that wouldn't be supported by NFS, since the place where the write happens is not the same kernel as the one where the notification needs to take place.

This is more generally motivation to resurrect the node-local, or process-local, compilation, which I will try and do in october.

dorugeber:
Does this act gracefully if compilation fails?  (e.g. the user hand-writes an `Expression` with broken syntax)

wence-:
https://github.com/OP2/PyOP2/pull/534/files#diff-ee93db0c87a2a587760fa1b0d7396108L174 (by default check_src_hashes is True) would catch that case.

tkarna:
Is this fix something that could be merged to PyOP2? Or do we wait for the more robust local compilation? I know that this fix isn't very elegant, and it does slow down compilation quite a bit on a high core count (but at least it works).

wence-:
Can you try #539?

wence-:
You need to set `PYOP2_NODE_LOCAL_COMPILATION=1` (default).  And you also need to set `PYOP2_CACHE_DIR` to a directory that points to a node-local filesystem.

tkarna:
Thanks, local compilation seems to work on our cluster. All runs have been successful thus far on ~100 cores.

wence-:
I did node-local compilation instead (#539).

PR #536, opened 2017-10-04
Handle gcc 7 dumpfullversion inanity

wence-:


inducer:
Confirming that this handles gcc 7 on Debian with 

```
$ gcc -dumpversion                                                                                                                                                                                       andreask_work@stout 11:37
7
$ gcc -dumpfullversion                                                                                                                                                                                   7.2.0
```

PR #538, opened 2017-10-20
De-unroll generated basis function loops

### LINKED ISSUE #537: Unrolled wrapper code is quite unhealthy for high-order kernels
In most cases, PyOP2 generates unrolled loops in the wrapper. With high-order elements, that generates a lot of code, and GCC fails to de-unroll them, so it becomes a lot of code binary as well.

This also makes the C compilation times minutes instead of seconds.

miklos1:
Resolves #537. Now quick compilation with GCC and small binary.

### DIFF
--- a/pyop2/sequential.py
+++ b/pyop2/sequential.py
@@ -33,7 +33,6 @@
 
 """OP2 sequential backend."""
 
-import builtins
 import os
 from textwrap import dedent
 from copy import deepcopy as dcopy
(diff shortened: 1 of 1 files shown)


miklos1:
Please review!

pyop2/sequential.py
```diff
@@ -98,6 +98,20 @@ def c_map_name(self, i, j):
     def c_offset_name(self, i, j):
         return self.c_arg_name() + "_off%d_%d" % (i, j)
 
+    def c_offset_decl(self):
+        maps = as_tuple(self.map, Map)
+        val = []
+        for i, map in enumerate(maps):
+            if not map.iterset._extruded:
```
tj-sun: consider _map i/o map?

PR #539, opened 2017-11-03
Support node-local compilation

wence-:
Do this by splitting the input communicators to the Compiler
constructor.  If MPI-3 is available, we use MPI_Split_type, otherwise
we rely on querying the filesystem.  The resulting communicator is
stashed as an attribute on the input communicator to the compile
command (freed when that communicator disappears).


wence-:
Any comments?

PR #545, opened 2017-12-06
value packing

tj-sun:
This PR fixes bugs in buffer size calculation when packing values instead of packing pointers.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3931,6 +3931,8 @@ class ParLoop(LazyComputation):
         self._iteration_region = kwargs.get("iterate", None)
         self._pass_layer_arg = kwargs.get("pass_layer_arg", False)
 
+        check_iterset(self.args, iterset)
+
         if self._pass_layer_arg:
             if self.is_direct:
                 raise ValueError("Can't request layer arg for direct iteration")
@@ -4146,6 +4148,38 @@ class ParLoop(LazyComputation):
         a certain part of an extruded mesh, for example on top cells, bottom cells or
         interior facets."""
         return self._iteration_region
+
+
+def check_iterset(args, iterset):
+    """Checks that the iteration set of the :class:`ParLoop` matches the
+    iteration set of all its arguments. A :class:`MapValueError` is raised
+    if this condition is not met.
+    Also determines the size of the local iteration space and checks all
+    arguments using an :class:`IterationIndex` for consistency."""
+
+    if isinstance(iterset, Subset):
+        _iterset = iterset.superset
+    else:
+        _iterset = iterset
+    if configuration["type_check"]:
+        if isinstance(_iterset, MixedSet):
+            raise SetTypeError("Cannot iterate over MixedSets")
+        for i, arg in enumerate(args):
+            if arg._is_global:
+                continue
+            if arg._is_direct:
+                if arg.data.dataset.set != _iterset:
+                    raise MapValueError(
+                        "Iterset of direct arg %s doesn't match ParLoop iterset." % i)
+                continue
+            for j, m in enumerate(arg._map):
+                if isinstance(_iterset, ExtrudedSet):
+                    if m.iterset != _iterset and m.iterset not in _iterset:
+                        raise MapValueError(
+                            "Iterset of arg %s map %s doesn't match ParLoop iterset." % (i, j))
+                elif m.iterset != _iterset and m.iterset not in _iterset:
+                    raise MapValueError(
+                        "Iterset of arg %s map %s doesn't match ParLoop iterset." % (i, j))
 
 
 @collective
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -4154,6 +4154,7 @@ def check_iterset(args, iterset):
     """Checks that the iteration set of the :class:`ParLoop` matches the
     iteration set of all its arguments. A :class:`MapValueError` is raised
     if this condition is not met.
+
     Also determines the size of the local iteration space and checks all
     arguments using an :class:`IterationIndex` for consistency."""
 
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -301,8 +301,6 @@ class Arg(object):
             self._block_shape = tuple(tuple((mr.arity, mc.arity)
                                       for mc in map[1])
                                       for mr in map[0])
-        elif self._uses_itspace:
-            self._block_shape = tuple(((m.arity,),) for m in map)
         else:
             self._block_shape = None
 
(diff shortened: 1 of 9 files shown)


pyop2/base.py
```diff
@@ -4327,7 +4327,7 @@ def build_itspace(args, iterset):
             if arg._uses_itspace:
                 _block_shape = arg._block_shape
                 if block_shape and block_shape != _block_shape:
-                    raise IndexValueError("Mismatching iteration space size for argument %d" % i)
+                    pass  # Allow different block shape
```
miklos1: So what is this thing with the mismatching block size?

pyop2/sequential.py
```diff
@@ -1096,20 +1096,20 @@ def extrusion_loop():
...
             # Readjust size to take into account the size of a vector space
-            _dat_size = (arg.data.cdim,)
-            _buf_size = [sum([e*d for e, d in zip(_buf_size, _dat_size)])]
-            _loop_size = [_buf_size[i]//_dat_size[i] for i in range(len(_buf_size))]
+            _dat_size = [_arg.data.cdim for _arg in arg]
+            _buf_size = [sum([e*d for e, d in zip(_loop_size, _dat_size)])]
         else:
             _dat_size = arg.data.dims[0][0]  # TODO: [0][0] ?
-            _buf_size = [e*d for e, d in zip(_buf_size, _dat_size)]
+            _buf_size = [e*d for e, d in zip(_loop_size, _dat_size)]
```
miklos1: Can you explain what is what here, and why does `_buf_size` change to `_loop_size`?

pyop2/sequential.py
```diff
@@ -1096,20 +1096,20 @@ def extrusion_loop():
...
         _buf_decl[arg] = arg.c_buffer_decl(_buf_size, count, _buf_name[arg], is_facet=is_facet)
         _tmp_decl[arg] = arg.c_buffer_decl(_buf_size, count, _tmp_name[arg], is_facet=is_facet,
                                            init=False)
+        facet_mult = 2 if is_facet else 1
         if arg.access not in [WRITE, INC]:
-            _itspace_loops = '\n'.join(['  ' * n + itspace_loop(n, e) for n, e in enumerate(_loop_size)])
+            _itspace_loops = '\n'.join(['  ' * n + itspace_loop(n, e*facet_mult) for n, e in enumerate(_loop_size)])
```
miklos1: Why did this work before?

pyop2/sequential.py
tj-sun: The buffer sizes seem wrong form many cases, e.g. `tests/regression/test_2dcohomology.py::test_betti1[space5]`

The generated wrapper is
```C
double buffer_arg0_0[20] __attribute__((aligned(16))) = {0.0};
double buffer_arg1_0[40] __attribute__((aligned(16)));
double buffer_arg2_0[20] __attribute__((aligned(16)));
double buffer_arg3_0[40] __attribute__((aligned(16)));
```
whereas the correct sizes are
```C
double buffer_arg0_0[20] __attribute__((aligned(16))) = {0.0};
double buffer_arg1_0[6] __attribute__((aligned(16)));
double buffer_arg2_0[20] __attribute__((aligned(16)));
double buffer_arg3_0[30] __attribute__((aligned(16)));
```
As far as I see, it seems to be a consequence of computing the buffer sizes from iteration extents.

pyop2/sequential.py
tj-sun: I think this was taken care of by `map_init` which has a flag `is_facet`

https://github.com/OP2/PyOP2/blob/master/pyop2/sequential.py#L1071

tj-sun:
This commit passes all tests except fusion.

The [`iterset_loop_body`](https://github.com/OP2/PyOP2/blob/value-packing-pr/pyop2/sequential.py#L1123) is a bit ugly as it was previously modeled on top of `IterationSpace`, but I think it handles mixed forms correctly now.

pyop2/sequential.py
```diff
@@ -843,13 +843,12 @@ def compile(self):
         # Blow away everything we don't need any more
         del self._args
         del self._kernel
-        del self._itspace
```
miklos1: You sure don't want to delete `self._iterset` instead?

miklos1:
Why does this change break fusion?

tj-sun:
It's just that I haven't updated the fusion code since removing `IterationSpace`, there are a few dependencies.

pyop2/sequential.py
tj-sun: Make sense. Amended.

miklos1:
> It's just that I haven't updated the fusion code since removing IterationSpace, there are a few dependencies.

Would that be as mechanical as the simple updates in `sequential.py`, or is there anything more challenging? I have an aversion to deliberately leaving code broken...

test/unit/test_indirect_loop.py
```diff
@@ -124,6 +124,7 @@ def test_mismatching_indset(self, iterset, x):
             op2.par_loop(op2.Kernel("", "dummy"), iterset,
                          x(op2.WRITE, op2.Map(iterset, op2.Set(nelems), 1)))
 
+    @pytest.mark.skip("allow different block size")
```
miklos1: Just remove this test, it does not apply anymore.

tj-sun:
I mainly needs to modify the fusion code, hopefully nothing needs to be done in `sequential.py`, so shouldn't be too bad (I think).

test/unit/test_indirect_loop.py
tj-sun: Cool.

miklos1:
> I mainly needs to modify the fusion code, hopefully nothing needs to be done in sequential.py, so shouldn't be too bad (I think).

OK, either that, or if we officially don't care about the fusion code, then file a separate PR that extirpates all of it.

pyop2/base.py
```diff
@@ -4272,39 +4149,18 @@ def iteration_region(self):
...
-def build_itspace(args, iterset):
-    """Creates an class:`IterationSpace` for the :class:`ParLoop` from the
-    given iteration set.
-
-    Also checks that the iteration set of the :class:`ParLoop` matches the
+def check_iterset(args, iterset):
+    """Checks that the iteration set of the :class:`ParLoop` matches the
     iteration set of all its arguments. A :class:`MapValueError` is raised
     if this condition is not met.
-
```
miklos1: This looks like the accidental removal of a healthy blank line (?)

pyop2/sequential.py
```diff
@@ -1132,45 +1131,50 @@ def itset_loop_body(i, j, shape, offsets, is_facet=False):
...
             elif arg._is_mat:
                 continue
             elif arg._is_dat:
-                loop_size = shape[0]*mult
-                _itspace_loops, _itspace_loop_close = itspace_loop(0, loop_size), '}'
-                _scatter_stmts = arg.c_buffer_scatter_vec(count, i, j, offsets, _buf_name[arg])
-                _buf_offset, _buf_offset_decl = '', ''
+                arg_scatter = []
+                offset = 0
+                for i, m in enumerate(arg.map):
```
miklos1: Is this loop to support mixed?

pyop2/base.py
tj-sun: Think so. Added back.

pyop2/sequential.py
tj-sun: Yes indeed.

PR #547, opened 2018-02-28
handles loopy kernel from tsfc

tj-sun:
We check if we get a loopy kernel from tsfc and generate code accordingly.


tj-sun:
Hi @miklos1 ,

My current plan is to go directly to `tsfc_loopying` branch, which generates the the wrapper as loopy kernel as well. It mostly works now but needs some cleaning up, so not quite ready for review yet - I think I need couple of weeks more on it.

Hope you doing well?

PR #551, opened 2018-07-26
Matis

wence-:
Add new support for even fancier PETSc matrix types.  Allows access to GENEO and BDDC preconditioners.

Not yet for block size > 1, but coming soon.


wence-:
Needs to land at the same time as firedrakeproject/firedrake#1354

sv2518:
We closed firedrakeproject/firedrake#1354, so we can close this as well I assume.

PR #552, opened 2018-08-21
loopy wrapper and kernel

tj-sun:
This pull request aim to
- support loopy kernel for parloops
- rewrite wrapper code generation in `sequential.py` (based on Lawrence's code)
- JITModule caching without code generation

This is not yet ready to go in until https://gitlab.tiker.net/inducer/loopy/merge_requests/246 is merged in loopy. As a result, the exact way to register callable kernel in loopy will need to be amended when that happens. But hopefully the overall structure should be immune to that update.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -47,7 +47,7 @@ import operator
 import types
 from hashlib import md5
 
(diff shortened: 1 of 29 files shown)


FabioLuporini:
basically we've got pyop3 now :smile:

PR #557, opened 2018-12-07
Apply boundary conditions through local to global maps

wence-:
Massive simplification of some really horrible code. Nearly ready I think. Matches up with firedrakeproject/firedrake#1348.


pyop2/base.py
```diff
@@ -3997,6 +3784,11 @@ def _jitmodule(self):
     def compute(self):
         """Executes the kernel over all members of the iteration space."""
         with timed_region("ParLoopExecute"):
+            lgmaps = []
+            for arg in self.args:
+                if arg._is_mat and arg.lgmaps is not None:
+                    lgmaps.append(arg.data.handle.getLGMap())
+                    arg.data.handle.setLGMap(*arg.lgmaps)
```
wence-: The only problem with this is that PETSc barfs if the block size changed. So I think that means when we make a *matrix* we have to say what boundary conditions we will need (should it be block or unwound) and use the appropriate block size there.

wence-:
If we like, we can wait until the loopy stuff is merged before going ahead with this. Or we can merge it if it's going to simplify the code. That is, once I fix the remaining issues.

wence-:
I think I will merge this after the loopy changes. Since I do as well need to get some changes into PETSc.

wence-:
This will arrive via a new branch once #559 is merged.

PR #558, opened 2019-01-23
Removed deprecated use of numpy.asscalar

danshapero:
This patch fixes warnings in numpy 1.16 that `asscalar` is deprecated in favor of the `item` method of ndarray.


wence-:
Thanks!

PR #559, opened 2019-04-04
loopy codegen

wence-:
Migrate to loopy-based code generation. Supersedes #555, let's review this one.


pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+
+    wrapper = loopy.make_kernel(domains,
+                                statements,
+                                kernel_data=parameters.kernel_data,
+                                target=loopy.CTarget(),
+                                temporary_variables=parameters.temporaries,
+                                symbol_manglers=[symbol_mangler],
+                                options=options,
+                                assumptions=assumptions,
+                                lang_version=(2018, 2),
```
wence-: Do we want `(2019, 1)` here now?

pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+    preamble = "\n".join(sorted(headers))
+
+    from coffee.base import Node
+
+    if isinstance(kernel._code, loopy.LoopKernel):
+        knl = kernel._code
+        wrapper = loopy.register_callable_kernel(wrapper, knl)
+        from loopy.transform.callable import _match_caller_callee_argument_dimension_
+        wrapper = _match_caller_callee_argument_dimension_(wrapper, knl.name)
+        wrapper = loopy.inline_callable_kernel(wrapper, knl.name)
```
wence-: @inducer/@kaushikcfd, do we always have to inline the kernel if it is a loopy kernel? Also, this matching of the argument dimensions, what do we need to change in the kernel generation that it is unneeded?

pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+    else:
+        # kernel is a string, add it to preamble
+        if isinstance(kernel._code, Node):
+            code = kernel._code.gencode()
+        else:
+            code = kernel._code
+        wrapper = loopy.register_function_id_to_in_knl_callable_mapper(
+            wrapper,
+            PyOP2KernelLookup(kernel.name, code, tuple(builder.argument_accesses)))
+        preamble = preamble + "\n" + code
```
wence-: OK, so this is basically "paste in C-string code".

pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+
+    wrapper = loopy.register_preamble_generators(wrapper, [_PreambleGen(preamble)])
+
+    # register petsc functions
+    wrapper = loopy.register_function_id_to_in_knl_callable_mapper(
+        wrapper, petsc_function_lookup)
```
wence-: Single line.

pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+
+
+@singledispatch
+def statement(expr, context):
+    raise AssertionError("Unhandled statement type '%s'" % type(expr))
+
+
+@statement.register(DummyInstruction)
+def statement_dummy(expr, context):
+    new_children = tuple(expression(c, context.parameters) for c in expr.children)
```
wence-: OK, so this just traverses the children to "register" the loop indices the dummy instruction contains.

pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+                            predicates=predicates,
+                            id=id,
+                            depends_on=depends_on, depends_on_is_final=True)
+
+
+@statement.register(FunctionCall)
+def statement_functioncall(expr, context):
+
+    from loopy.symbolic import SubArrayRef
+    from loopy.types import OpaqueType
```
wence-: imports at top-level, and fewer blank spaces.

pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+            sweeping_indices = []
+            for index in var.index_tuple:
+                indices.append(index)
+                if isinstance(index, pym.Variable) and index.name in free_indices:
+                    sweeping_indices.append(index)
+            arg = SubArrayRef(tuple(sweeping_indices), var)
+        else:
+            # scalar argument or constant
+            arg = var
+        if access is READ or (isinstance(child, Argument) and isinstance(child.dtype, OpaqueType)):
```
wence-: OK, so for `OpaqueType`s (pointers), we just have to lie to loopy about any actual access descriptor and just say `READ`.

pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+    @translate.register(Comparison)
+    def translate_comparison(expr, vars):
+        a, b = (translate(c, vars) for c in expr.children)
+        fn = {">": "gt_set",
+              ">=": "ge_set",
+              "==": "eq_set",
+              "!=": "ne_set",
+              "<": "lt_set",
+              "<=": "le_set"}[expr.operator]
+        return getattr(a, fn)(b)
```
wence-: More complicated constraints on indices would require extending this.

pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+
+@expression.register(MultiIndex)
+def expression_multiindex(expr, parameters):
+    return tuple(expression(c, parameters) for c in expr.children)
+
+
+@expression.register(Extent)
+def expression_extent(expr, parameters):
+    multiindex, = expr.children
+    return int(numpy.prod(tuple(i.extent for i in multiindex)))
```
wence-: This is wrong for vectorisable functions taking extent indices that will be vectorised (because we encode the extent in the loopy statements).

pyop2/compilation.py
```diff
@@ -409,40 +422,50 @@ def __init__(self, cppargs=[], ldargs=[], cpp=False, comm=None):
...
     :arg extension: extension of the source file (c, cpp)
     :arg fn_name: The name of the function to return from the resulting library
     :arg cppargs: A list of arguments to the C compiler (optional)
     :arg ldargs: A list of arguments to the linker (optional)
-    :arg argtypes: A list of ctypes argument types matching the
-         arguments of the returned function (optional, pass ``None``
-         for ``void``).
+    :arg argtypes: A list of ctypes argument types matching the arguments of
+         the returned function (optional, pass ``None`` for ``void``). This is
+         only used when string is passed in instead of JITModule.
```
wence-: So the reason for this code churn is that the code can now either be something that makes code (via wrapper generation), or the string itself. So perhaps we should pass in an object that can return a cache key and the code in both cases. Then this change would be smaller.


e.g.
```
class StrCode(object):
    def __init__(self, code):
       self.code_to_compile = code
    @cached_property
    def cache_key(self):
       return md5(self.code_to_compile.encode())
```

pyop2/op2.py
```diff
@@ -62,6 +61,9 @@
 
 _initialised = False
 
+# turn off loopy caching because pyop2 kernels are cached already
+loopy.set_caching_enabled(False)
```
wence-: OK, so we don't need to do this again in Firedrake.

pyop2/codegen/rep2loopy.py
kaushikcfd: 1. No, if the inline transformation is not called then separate functions are generated and the corresponding arguments are passed.
2. As far as I remember, there is a mismatch in the dimensions of the "coordinates" argument between the caller and callee(for ex. in tsfc it may have shape`(6, )` and in pyop2 `(3, 2)`) . And that's why we need `_match_caller_callee_argument_dimension_` which changes the shape of the arguments of the callee kernel assuming C-ordering of the data layout.

wence-: OK, so if I remove the `inline_callable_kernel` line, then everything should continue to work?

kaushikcfd: Yes.

wence-: I get:
```python
    def generate_code_v2(program):
        """
        Returns an instance of :class:`CodeGenerationResult`.
    
        :param program: An instance of :class:`loopy.Program`.
        """
        from loopy.kernel import LoopKernel
        from loopy.program import make_program
    
        if isinstance(program, LoopKernel):
            program = make_program(program)
    
        from loopy.kernel import KernelState
        if program.root_kernel.state == KernelState.INITIAL:
            from loopy.preprocess import preprocess_program
            program = preprocess_program(program)
    
        from loopy.type_inference import infer_unknown_types
        program = infer_unknown_types(program, expect_completion=True)
    
        codegen_results = {}
    
        for func_id, in_knl_callable in program.callables_table.items():
            if isinstance(in_knl_callable, CallableKernel):
                codegen_results[func_id] = (
                        generate_code_for_a_single_kernel(in_knl_callable.subkernel,
                            program.callables_table, program.target))
                if not in_knl_callable.subkernel.is_called_from_host:
                    assert codegen_results[func_id].host_program is None
    
        device_preambles = []
        for cgr in codegen_results.values():
            device_preambles.extend(cgr.device_preambles)
    
        # collecting the function declarations of callee kernels
        for in_knl_callable in program.callables_table.values():
            for preamble in in_knl_callable.generate_preambles(program.target):
                device_preambles.append(preamble)
    
        collective_device_program = codegen_results[program.name].device_programs[0]
        callee_fdecls = []
    
        for func_id, callee_cgr in codegen_results.items():
            if func_id != program.name:
                assert len(callee_cgr.device_programs) == 1
                callee_prog_ast = callee_cgr.device_programs[0].ast
                collective_device_program = collective_device_program.copy(
                        ast=Collection([callee_prog_ast, collective_device_program.ast]))
>               callee_fdecls.append(callee_prog_ast.fdecl)
E               AttributeError: 'Collection' object has no attribute 'fdecl'

../loopy/loopy/codegen/__init__.py:623: AttributeError```

It looks like this code assumes that the kernel has a single thing which is a functionbody, but that doesn't handle the case we have where we have a `Collection`.

If I try the obvious patch:
```
diff --git a/loopy/codegen/__init__.py b/loopy/codegen/__init__.py
index 8f3e15f2..1e5ff8b1 100644
--- a/loopy/codegen/__init__.py
+++ b/loopy/codegen/__init__.py
@@ -620,7 +620,14 @@ def generate_code_v2(program):
             callee_prog_ast = callee_cgr.device_programs[0].ast
             collective_device_program = collective_device_program.copy(
                     ast=Collection([callee_prog_ast, collective_device_program.ast]))
-            callee_fdecls.append(callee_prog_ast.fdecl)
+            if isinstance(callee_prog_ast, Collection):
+                for entry in callee_prog_ast.contents:
+                    try:
+                        callee_fdecls.append(entry.fdecl)
+                    except AttributeError:
+                        pass
+            else:
+                callee_fdecls.append(entry.fdecl)
 
     # collecting the function declarations of callee kernels
     for callee_fdecl in callee_fdecls:
```

then I get a further error (somehow some arguments with shape (1, ) have been rewritten to shape ().

I have an inner kernel:
```
KERNEL: expression
ARGUMENTS:
fn_0: type: np:dtype('float64'), shape: (1), dim_tags: (N0:stride:1) aspace: global
fn_1: type: np:dtype('float64'), shape: (1), dim_tags: (N0:stride:1) aspace: global
fn_2: type: np:dtype('float64'), shape: (1), dim_tags: (N0:stride:1) aspace: global
DOMAINS:
{ [d] : d = 0 }
INAME IMPLEMENTATION TAGS:
d: None
INSTRUCTIONS:
for d
  fn_0[d] = fn_1[d] + (-1)*fn_2[d]  {id=insn}
end d
------
```

And an outer kernel:
```
KERNEL: wrap_expression
ARGUMENTS:
dat0: type: np:dtype('float64'), shape: (None), dim_tags: (N0:stride:1) aspace: global
dat1: type: np:dtype('float64'), shape: (None), dim_tags: (N0:stride:1) aspace: global
dat2: type: np:dtype('float64'), shape: (None), dim_tags: (N0:stride:1) aspace: global
end: ValueArg, type: np:dtype('int32')
start: ValueArg, type: np:dtype('int32')
DOMAINS:
[end, start] -> { [n] : start <= n < end }
INAME IMPLEMENTATION TAGS:
n: None
INSTRUCTIONS:
for n
  SubArrayRef((), (dat0[n])) = expression(SubArrayRef((), (dat1[n])), SubArrayRef((), (dat2[n])))  {id=statement0}
end n
```
```
But I get an error:

            if len(subscript) != len(shape):
                raise LoopyError("subscript to '%s' in '%s' has the wrong "
                        "number of indices (got: %d, expected: %d)" % (
                            expr.aggregate.name, expr,
>                           len(subscript), len(shape)))
E               loopy.diagnostic.LoopyError: subscript to 'fn_0' in 'fn_0[d]' has the wrong number of indices (got: 1, expected: 0)
```

And indeed, something has rewritten the inner kernel into:
```
KERNEL: expression
ARGUMENTS:
fn_0: type: np:dtype('float64'), shape: () aspace: global
fn_1: type: np:dtype('float64'), shape: () aspace: global
fn_2: type: np:dtype('float64'), shape: () aspace: global
DOMAINS:
{ [d] : d = 0 }
INAME IMPLEMENTATION TAGS:
d: None
INSTRUCTIONS:
for d
  fn_0[d] = fn_1[d] + (-1)*fn_2[d]  {id=insn}
end d```

kaushikcfd: Let me fix it.

pyop2/codegen/builder.py
```diff
@@ -0,0 +1,833 @@
...
+        return self._pack
+
+    def kernel_arg(self, loop_indices=None):
+        if self.map_ is None:
+            if loop_indices is None:
+                raise ValueError("Need iteration index")
+            n, layer = self.pick_loop_indices(*loop_indices)
+            # Direct dats on extruded sets never get a layer index
+            # (they're defined on the "base" set, effectively).
+            # FIXME: is this a bug?
```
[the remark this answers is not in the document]
wence-: Removed comment.

pyop2/codegen/builder.py
```diff
@@ -0,0 +1,833 @@
...
+            args = args + (self.layer_index, )
+            access = access + (READ,)
+        if self.forward_arguments:
+            args = self.forward_arguments + args
+            access = tuple([WRITE] * len(self.forward_arguments)) + access
+        return FunctionCall(self.kernel.name, KernelInst(), access, free_indices, *args)
+
+    def emit_instructions(self):
+        loop_indices = [x for x in self.loop_indices if x is not None]
+        yield DummyInstruction(PackInst(), *loop_indices)
```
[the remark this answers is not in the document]
wence-: Done.

pyop2/codegen/rep2loopy.py
```diff
@@ -0,0 +1,781 @@
...
+        import re
+        pattern = re.compile(r"^([a-zA-Z_]+)([0-9]+$)")
+        replace = {}
+        names = defaultdict(list)
+        for node in traversal(instructions):
+            if isinstance(node, (Index, RuntimeIndex, Variable, Argument)):
+                match = pattern.match(node.name)
+                if match is not None:
+                    prefix, idx = match.groups()  # string, index
+                    names[prefix].append(int(idx))
```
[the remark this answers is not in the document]
wence-: I did this.

pyop2/codegen/rep2loopy.py
wence-: Thanks!

pyop2/codegen/rep2loopy.py
wence-: Added comment.

pyop2/codegen/builder.py
```diff
@@ -0,0 +1,833 @@
...
+            args.append(map_.values)
+        return tuple(args)
+
+    def kernel_call(self):
+        args = self.kernel_args
+        access = tuple(self.argument_accesses)
+        # assuming every index is free index
+        free_indices = set(itertools.chain.from_iterable(arg.multiindex for arg in args))
+        # remove runtime index
+        free_indices = tuple(i for i in free_indices if isinstance(i, Index))
```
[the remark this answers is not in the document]
wence-: No, they are only use to check which indices are part of an array reference later.

pyop2/codegen/rep2loopy.py
wence-: @inducer / @kaushikcfd any thoughts?

kaushikcfd: @inducer is more appropriate to answer this one.

inducer: There's no `(2019, 1)` version, at least as yet. And until we need to break compatibility in some way, the language version will stay at `(2018, 2)`.

wence-: OK, thanks.

pyop2/codegen/rep2loopy.py
kaushikcfd: - The problem was the support of empty subarray refs. That has been fixed and pushed to `firedrakeproject/loopy`.
- Corresponding test: https://gitlab.tiker.net/inducer/loopy/blob/firedrake/test/test_callables.py#L449 (tests both with and without inline)

wence-: OK, this stuff is now fixed, thanks!

wence-:
Thanks all!

PR #561, opened 2019-04-30
codegen: Fix non-determinism in assumptions ordering

wence-:
loopy.assume doesn't necessarily maintain order across processes, so
do it "by hand".


wence-:
@tkarna I think this fixes your point eval issue.

tkarna:
Thanks a bunch! This indeed fixes the issue (upgrading python version had no effect).

Issue #564, opened 2019-06-13
`as_tuple` gives uninformative error messages on type error


danshapero:
I'm getting an error [here](https://github.com/OP2/PyOP2/blob/3c88f9e7ef572387d8a78ee863b79349bc918272/pyop2/utils.py#L86) that happened from passing a numpy int32 where an int was expected. A numpy int32 is rejected because it is not an instance of plain old int. The error message didn't make this clear.

This error was triggered when passing a tuple of subdomain IDs to the measure `ds` in order to integrate over only part of the boundary. The IDs came from a call to `mesh.exterior_facets.unique_markers`, which returns a numpy array of numpy.int32 objects. So this error could be especially easy to trigger if users are getting subdomain IDs by querying the mesh.

danshapero:
Fixed by firedrakeproject/firedrake#1454.

PR #565, opened 2019-06-14
Make `as_tuple` more permissive

danshapero:
This patch makes the function `utils.as_tuple` accept an iterable of types to check the input item against. That way it can be called to check whether the item members are either an int, a numpy int32, or a numpy int64, rather than just an int. This can address #564 by making the error case less likely.

### DIFF
--- a/pyop2/utils.py
+++ b/pyop2/utils.py
@@ -63,7 +63,7 @@ class cached_property(object):
         return result
 
 
-def as_tuple(item, type=None, length=None, allow_none=False):
+def as_tuple(item, typs=None, length=None, allow_none=False):
     # Empty list if we get passed None
     if item is None:
         t = ()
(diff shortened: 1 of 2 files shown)


pyop2/utils.py
```diff
@@ -77,13 +77,16 @@ def as_tuple(item, type=None, length=None, allow_none=False):
...
+            try:
+                typs = tuple(typs)
+            except TypeError:
+                typs = tuple((typs,))
             if allow_none:
-                valid = all((isinstance(i, type) or i is None) for i in t)
-            else:
-                valid = all(isinstance(i, type) for i in t)
+                typs += (type(None),)
+            valid = all(any(isinstance(i, typ) for typ in typs) for i in t)
```
wence-: `isinstance` accepts an iterable of things to check. So even without this patch, `as_tuple` allows one to check for more than one type.

```python
In [1]: from pyop2.utils import as_tuple

In [2]: as_tuple([1.54, 1], int)
TypeError                                 Traceback (most recent call last)
<ipython-input-2-072f65d2b25a> in <module>
----> 1 as_tuple([1.54, 1], int)

~/tmp/firedrake/src/PyOP2/pyop2/utils.py in as_tuple(item, type, length, allow_none)
     84                 valid = all(isinstance(i, type) for i in t)
     85             if not valid:
---> 86                 raise TypeError("Items need to be of type %s" % type)
     87     return t
     88 

TypeError: Items need to be of type <class 'int'>

In [3]: as_tuple([1.54, 1], (int, float))
Out[3]: (1.54, 1)
```

danshapero: Huh, I even [used that feature myself](https://github.com/icepack/icepack/blob/master/icepack/interpolate.py#L35), I just forgot you could do that! I'll close this PR then, the issue I was having can be fixed by changing the calls within firedrake.

Issue #566, opened 2019-07-12
setup.py not installing codegen


JaroslavHron:
doing direct installation into prefix directory 
`python3 -s setup.py --no-user-cfg install --prefix=/usr/local/pkg/pyop2`
will not install the `codegen` subpackage

it has to be explicitly listed as
```
setup( ..., packages=['pyop2','pyop2.codegen'], ....)
```
or with `setuptools` you coud use `find_packages()`
```
from setuptools import setup, find_packages

setup( ..., packages=find_packages(), ....)
```

wence-:
Merged your pr. thanks.

PR #567, opened 2019-07-12
update setup.py to install codegen subpackage

JaroslavHron:
`codegen` subpackage has to be explicitly mentioned in setup to be installed - see #566


wence-:
Ah, sorry this has been languishing so long, thanks!

PR #573, opened 2019-12-19
utils: Backport cached property from functools

wence-:
Fixes some corner cases of decorator composition with cached_property (c.f. firedrakeproject/firedrake#1567)

### DIFF
--- a/pyop2/utils.py
+++ b/pyop2/utils.py
@@ -43,24 +43,49 @@ import argparse
 from pyop2.exceptions import DataTypeError, DataValueError
 from pyop2.configuration import configuration
 
+try:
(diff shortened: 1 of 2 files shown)


wence-:
OK, this is more complicated, because it requires Python 3.6. (released Dec 2016). I guess that's a bit recent (we presently require 3.5).

nbouziani:
Could we not just let the hacking name workaround?

sv2518:
Looks like firedrakeproject/firedrake#1567 already got merged, is this merge request still active?

wence-:
Probably not. Sorry for the delay.

PR #575, opened 2020-03-13
Fix MixedMap._kernel_args_ with R block

pefarrell:
Getting the `_kernel_args_` of a `MixedMap` crashes if the mixed function space features an R block. This commit fixes this.

Question for reviewers: should the same `if m is not None` be added everywhere else in `MixedMap`? I'm happy to do this if you agree it's the right thing.


wence-:
Yes, I think you need the same thing for `_argtypes_` and `_wrapper_cache_key_`.

pefarrell:
OK, done.

PR #578, opened 2020-04-23
Support mat_type dense

pefarrell:
Dense matrices are useful for e.g. spectral methods.

### DIFF
--- a/pyop2/petsc_base.py
+++ b/pyop2/petsc_base.py
@@ -601,6 +601,7 @@ class Mat(base.Mat):
     for each element in the :class:`Sparsity`."""
 
     def __init__(self, *args, **kwargs):
+        self.mat_type = kwargs.pop("mat_type", None)
         base.Mat.__init__(self, *args, **kwargs)
         self._init()
         self.assembly_state = Mat.ASSEMBLED
@@ -617,8 +618,10 @@ class Mat(base.Mat):
         if not self.dtype == PETSc.ScalarType:
             raise RuntimeError("Can only create a matrix of type %s, %s is not supported"
                                % (PETSc.ScalarType, self.dtype))
+        if self.mat_type == "dense":
+            self._init_dense()
         # If the Sparsity is defined on MixedDataSets, we need to build a MatNest
-        if self.sparsity.shape > (1, 1):
+        elif self.sparsity.shape > (1, 1):
             if self.sparsity.nested:
                 self._init_nest()
                 self._nested = True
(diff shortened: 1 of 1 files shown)


pyop2/petsc_base.py
```diff
@@ -601,6 +601,7 @@ class Mat(base.Mat):
     for each element in the :class:`Sparsity`."""
 
     def __init__(self, *args, **kwargs):
+        self.mat_type = kwargs.pop("mat_type", None)
```
pefarrell, line 604: I found it strange that the OP2 Mat class didn't know the `mat_type` before.

In the future, it could be useful to inspect this for e.g. the aijmkl class from MKL.

pyop2/petsc_base.py
```diff
@@ -627,6 +630,31 @@ def _init(self):
...
+        self._blocks = []
+        rows, cols = self.sparsity.shape
+        for i in range(rows):
+            row = []
+            for j in range(cols):
+                row.append(MatBlock(self, i, j))
+            self._blocks.append(row)
+        mat.setOption(mat.Option.IGNORE_OFF_PROC_ENTRIES, False)
+        mat.setOption(mat.Option.SUBSET_OFF_PROC_ENTRIES, True)
+        mat.setUp()
```
pefarrell, line 652: The call to `setUp` is necessary, otherwise the `zeroEntries` segfaults.

pyop2/petsc_base.py
```diff
@@ -627,6 +630,31 @@ def _init(self):
...
             self._init_block()
 
+    def _init_dense(self):
+        mat = PETSc.Mat()
+        rset, cset = self.sparsity.dsets
+        rlgmap = rset.unblocked_lgmap
+        clgmap = cset.unblocked_lgmap
+        mat.createDense(size=((self.nrows, None), (self.ncols, None)),
+                        bsize=1,
+                        comm=self.comm)
```
wence-, line 640: I'm not sure this is right, since now your matrix will only support `MatSetValues` not `MatSetValuesBlocked` (and the decision we make as to whether to use one or the other depends on whether the maps are blocked, not the matrix).

Does it work if for assembling `VectorFunctionSpace` problems?

pefarrell: I added a test on the firedrake PR, comparing `aij` and `dense` for a vector-valued problem. Works fine.

What do you think would be safer?

pyop2/petsc_base.py
wence-: I guess ideally we would `setfromoptions` the petsc Mat under the hood. The issue is on the pyop2 side we need to do different things depending on the class of matrix. Hence all this special-casing...

pyop2/petsc_base.py
wence-: Ah, now I understand, because you make it with `MatBlock` things, everything falls out right, because the matblocks have the right block size.

PR #580, opened 2020-05-11
Inline callables from loopy programs

sv2518:
This is an attempt of handling multiple kernel as callables stacked into each other. The previous workaround for this was to generate c-code first and then attach the kernel to the pyop2 wrapper kernel, see https://github.com/firedrakeproject/firedrake/blob/418eaea8e307de39983cf9eb60caa8e01d266212/firedrake/slate/slac/compiler.py#L153.

When inlining the TSFC kernel in loopy, the Slate wrapper kernel becomes a Program.

In this PR I attempt to properly inline the Slate kernel in PyOP2. The idea is to first copy or register all resolved functions from the Slate wrapper program to the PyOP2 wrapper program (Copy for example for math functions, register for CallableKernels) and afterwards inline the Slate wrapper Kernel.

I ran into trouble with undefined keys in the history of Programs and a couple of other problems which are addressed in https://github.com/firedrakeproject/loopy/compare/allow-callablestables-gen-with-functionsnotinhistory.

I am not a 100percent sure, that this would be sufficient to inline the Slate wrapper. Does anyone have any comment on this?

### DIFF
--- a/pyop2/codegen/rep2loopy.py
+++ b/pyop2/codegen/rep2loopy.py
@@ -629,6 +629,27 @@ def generate(builder, wrapper_name=None):
         knl = kernel._code
         wrapper = loopy.register_callable_kernel(wrapper, knl)
         wrapper = _match_caller_callee_argument_dimension_(wrapper, knl.name)
+        wrapper = loopy.inline_callable_kernel(wrapper, knl.name)
+    elif isinstance(kernel._code, loopy.Program):
+        kernel = kernel._code  # loopy.Program
+        knl = kernel.root_kernel  # loopy.LoopKernel
+
+        from loopy.transform.callable import _match_caller_callee_argument_dimension_
+
+        # Register all resolved functions of root_kernel in wrapper
+        for name, callable in kernel.callables_table.resolved_functions.items():
+            if isinstance(callable, loopy.CallableKernel):
+                wrapper = loopy.register_callable_kernel(wrapper, callable.subkernel)
+            else:
+                # Mathcallables e.g. do not have subkernels so add by hand
+                combined_callables = wrapper.callables_table.resolved_functions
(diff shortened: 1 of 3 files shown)


pyop2/base.py
```diff
@@ -1689,9 +1690,9 @@ def _iop(self, other, op):
             self._check_shape(other)
             rhs = _other.index(i)
         insn = loopy.Assignment(lhs, op(lhs, rhs), within_inames=frozenset(["i"]))
-        data = [loopy.GlobalArg("self", dtype=self.dtype, shape=(self.cdim,)),
-                loopy.GlobalArg("other", dtype=other.dtype, shape=(other.cdim,))]
-        knl = loopy.make_function([domain], [insn], data, name=name)
+        data = [loopy.GlobalArg("self", dtype=self.dtype, shape=(self.cdim,), target=TARGET),
+                loopy.GlobalArg("other", dtype=other.dtype, shape=(other.cdim,), target=TARGET)]
+        knl = loopy.make_function([domain], [insn], data, name=name, target=TARGET)
```
wence-, line 1695: `make_function` doesn't take `target` I think.

sv2518: We specify one in TSFC as well https://github.com/firedrakeproject/tsfc/blob/d2fc040967c3c111f01f3a5c78239b5182073f7a/tsfc/loopy.py#L153 and the loopy doc agrees https://documen.tician.de/loopy/ref_creation.html.

wence-: Ah, OK, but then why only for this kernel and not the others?

sv2518:
I think this is not needed anymore since the functionality dropped out of adapting PyOP2 and Firedrake to the new Callable structures in Loo.py in the sprint earlier this year. I will close the PR.

PR #581, opened 2020-05-12
Introduce inverse and solve callable in PyOP2.

sv2518:
We need these to resolve inverse and solve calls at loopy level.

### DIFF
--- a/pyop2/codegen/rep2loopy.py
+++ b/pyop2/codegen/rep2loopy.py
@@ -3,7 +3,11 @@ import numpy
 
 import loopy
 from loopy.symbolic import SubArrayRef
-from loopy.types import OpaqueType
+from loopy.expression import dtype_to_type_context
+from pymbolic.mapper.stringifier import PREC_NONE
+from pymbolic import var
+from loopy.types import NumpyType, OpaqueType
+import abc
 
 import islpy as isl
 import pymbolic.primitives as pym
@@ -83,6 +87,163 @@ def petsc_function_lookup(target, identifier):
     if identifier in petsc_functions:
         return PetscCallable(name=identifier)
     return None
+
+
+class LACallable(loopy.ScalarCallable, metaclass=abc.ABCMeta):
+    """
+    The LACallable (Linear algebra callable)
+    replaces loopy.CallInstructions to linear algebra functions
+    like solve or inverse by LAPACK calls.
+    """
+    def __init__(self, name, arg_id_to_dtype=None,
+                 arg_id_to_descr=None, name_in_target=None):
+
+        super(LACallable, self).__init__(name,
+                                         arg_id_to_dtype=arg_id_to_dtype,
+                                         arg_id_to_descr=arg_id_to_descr)
+        self.name = name
+        self.name_in_target = name_in_target if name_in_target else name
+
+    @abc.abstractmethod
+    def generate_preambles(self, target):
+        pass
+
+    def with_types(self, arg_id_to_dtype, kernel, callables_table):
+        dtypes = OrderedDict()
+        for i in range(len(arg_id_to_dtype)):
+            if arg_id_to_dtype.get(i) is None:
+                # the types provided aren't mature enough to specialize the
+                # callable
+                return (self.copy(arg_id_to_dtype=arg_id_to_dtype),
+                        callables_table)
+            else:
+                mat_dtype = arg_id_to_dtype[i].numpy_dtype
+                dtypes[i] = NumpyType(mat_dtype)
+        dtypes[-1] = NumpyType(dtypes[0].dtype)
+
+        return (self.copy(name_in_target=self.name_in_target,
+                arg_id_to_dtype=dtypes),
+                callables_table)
+
+    def emit_call_insn(self, insn, target, expression_to_code_mapper):
+        assert self.is_ready_for_codegen()
+        assert isinstance(insn, loopy.CallInstruction)
+
+        parameters = insn.expression.parameters
+
+        parameters = list(parameters)
+        par_dtypes = [self.arg_id_to_dtype[i] for i, _ in enumerate(parameters)]
+
+        parameters.append(insn.assignees[-1])
+        par_dtypes.append(self.arg_id_to_dtype[0])
+
+        mat_descr = self.arg_id_to_descr[0]
+        arg_c_parameters = [
+            expression_to_code_mapper(
+                par,
+                PREC_NONE,
+                dtype_to_type_context(target, par_dtype),
+                par_dtype
+            ).expr
+            for par, par_dtype in zip(parameters, par_dtypes)
+        ]
+        c_parameters = [arg_c_parameters[-1]]
+        c_parameters.extend([arg for arg in arg_c_parameters[:-1]])
+        c_parameters.append(numpy.int32(mat_descr.shape[1]))  # n
+        return var(self.name_in_target)(*c_parameters), False
+
+
+class INVCallable(LACallable):
+    """
+    The InverseCallable replaces loopy.CallInstructions to "inverse"
+    functions by LAPACK getri.
+    """
+    def generate_preambles(self, target):
+        assert isinstance(target, loopy.CTarget)
+        inverse_preamble = """
+            #define Inverse_HPP
+            #define BUF_SIZE 30
+
+            static PetscBLASInt ipiv_buffer[BUF_SIZE];
+            static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
+            static void inverse(PetscScalar* __restrict__ Aout, const PetscScalar* __restrict__ A, PetscBLASInt N)
+            {
+                PetscBLASInt info;
+                PetscBLASInt *ipiv = N <= BUF_SIZE ? ipiv_buffer : malloc(N*sizeof(*ipiv));
+                PetscScalar *Awork = N <= BUF_SIZE ? work_buffer : malloc(N*N*sizeof(*Awork));
+                memcpy(Aout, A, N*N*sizeof(PetscScalar));
+                LAPACKgetrf_(&N, &N, Aout, &N, ipiv, &info);
+                if(info == 0){
+                    LAPACKgetri_(&N, Aout, &N, ipiv, Awork, &N, &info);
+                }
+                if(info != 0){
+                    fprintf(stderr, \"Getri throws nonzero info.\");
+                    abort();
+                }
+                if ( N > BUF_SIZE ) {
+                    free(Awork);
+                    free(ipiv);
+                }
+            }
+        """
+        yield ("inverse", "#include <petscsys.h>\n#include <petscblaslapack.h>\n" + inverse_preamble)
+        return
+
+
+def inv_fn_lookup(target, identifier):
+    if identifier == 'inv':
+        return INVCallable(name='inverse')
+    else:
+        return None
+
+
+class SolveCallable(LACallable):
+    """
+    The SolveCallable replaces loopy.CallInstructions to "solve"
+    functions by LAPACK getrs.
+    """
+    def generate_preambles(self, target):
+        assert isinstance(target, loopy.CTarget)
+        code = """
+            #define Solve_HPP
+            #define BUF_SIZE 30
+
+            static PetscBLASInt ipiv_buffer[BUF_SIZE];
+            static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
+            static void solve(PetscScalar* __restrict__ out, const PetscScalar* __restrict__ A, const PetscScalar* __restrict__ B, PetscBLASInt N)
+            {
+                PetscBLASInt info;
+                PetscBLASInt *ipiv = N <= BUF_SIZE ? ipiv_buffer : malloc(N*sizeof(*ipiv));
+                memcpy(out,B,N*sizeof(PetscScalar));
+                PetscScalar *Awork = N <= BUF_SIZE ? work_buffer : malloc(N*N*sizeof(*Awork));
+                memcpy(Awork,A,N*N*sizeof(PetscScalar));
+                PetscBLASInt NRHS = 1;
+                const char T = 'T';
+                LAPACKgetrf_(&N, &N, Awork, &N, ipiv, &info);
+                if(info == 0){
+                    LAPACKgetrs_(&T, &N, &NRHS, Awork, &N, ipiv, out, &N, &info);
+                }
+                if(info != 0){
+                    fprintf(stderr, \"Gesv throws nonzero info.\");
+                    abort();
+                }
+
+                if ( N > BUF_SIZE ) {
+                    free(ipiv);
+                    free(Awork);
+                }
+            }
+        """
+
+        yield ("solve", "#include <petscsys.h>\n#include <petscblaslapack.h>\n" + code)
+        return
+
+
+def solve_fn_lookup(target, identifier):
+    if identifier == 'solve':
+        return SolveCallable(name='solve')
+    else:
+        return None
 
 
 class _PreambleGen(ImmutableRecord):
--- a/test/unit/test_callables.py
+++ b/test/unit/test_callables.py
@@ -1,0 +1,121 @@
+# This file is part of PyOP2
+#
+# PyOP2 is Copyright (c) 2012-2014, Imperial College London and
+# others. Please see the AUTHORS file in the main source directory for
+# a full list of copyright holders.  All rights reserved.
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions
+# are met:
+#
+#     * Redistributions of source code must retain the above copyright
+#       notice, this list of conditions and the following disclaimer.
+#     * Redistributions in binary form must reproduce the above copyright
+#       notice, this list of conditions and the following disclaimer in the
+#       documentation and/or other materials provided with the distribution.
+#     * The name of Imperial College London or that of other
+#       contributors may not be used to endorse or promote products
+#       derived from this software without specific prior written
+#       permission.
+#
+# THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTERS
+# ''AS IS'' AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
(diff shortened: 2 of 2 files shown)


pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+                if(info!=0)
+                    fprintf(stderr,\"Getri throws nonzero info.\");
+                if ( N > BUF_SIZE ) {
+                    free(Awork);
+                    free(ipiv);
+                }
+            }
+            #endif
+        """
+        yield("inverse", "#include <petscsys.h>\n#include <petscblaslapack.h>\n"+inverse_preamble)
```
wence-, line 186: style, space before `(`.

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+        return
```
wence-, line 187: No need for `return`.

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+            {
+                PetscBLASInt info;
+                PetscBLASInt *ipiv = N <= BUF_SIZE ? ipiv_buffer : malloc(N*sizeof(*ipiv));
+                PetscScalar *Awork = N <= BUF_SIZE ? work_buffer : malloc(N*N*sizeof(*Awork));
+                memcpy(Aout,A,N*N*sizeof(PetscScalar));
+                LAPACKgetrf_(&N,&N,Aout,&N,ipiv,&info);
+                if(info==0)
+                    LAPACKgetri_(&N,Aout,&N,ipiv,Awork,&N,&info);
+                if(info!=0)
+                    fprintf(stderr,\"Getri throws nonzero info.\");
```
wence-, line 178: I think since you're wrapped in `"""` you don't need `\` here.

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+    def with_types(self, arg_id_to_dtype, kernel, callables_table):
+        for i in range(len(arg_id_to_dtype)):
+            if arg_id_to_dtype.get(i) is None:
+                # the types provided aren't mature enough to specialize the
+                # callable
+                return (self.copy(arg_id_to_dtype=arg_id_to_dtype),
+                        callables_table)
+
+        mat_dtype = arg_id_to_dtype[0].numpy_dtype
+        name_in_target = "solve"
```
wence-, line 217: This ignores `self.name_in_target` is that deliberate?

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+                dtype_to_type_context(target, par_dtype),
+                par_dtype
+            ).expr
+            if isinstance(par, SubArrayRef) else
+            expression_to_code_mapper(
+                par,
+                PREC_NONE,
+                dtype_to_type_context(target, par_dtype),
+                par_dtype
+            ).expr
```
wence-, line 250: The true and false branches to this ternary appear to be identical?

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+            for par, par_dtype in zip(parameters, par_dtypes)
+        ]
+        c_parameters = []
+        c_parameters.insert(0, arg_c_parameters[-1])  # out
+        c_parameters.insert(1, arg_c_parameters[0])  # A
+        c_parameters.insert(2, arg_c_parameters[1])  # B
+        c_parameters.insert(3, mat_descr_A.shape[1])  # n
```
wence-, line 257: Prefer `append` over `insert` at the end.

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+                PREC_NONE,
+                dtype_to_type_context(target, par_dtype),
+                par_dtype
+            ).expr
+            if isinstance(par, SubArrayRef) else
+            expression_to_code_mapper(
+                par,
+                PREC_NONE,
+                dtype_to_type_context(target, par_dtype),
+                par_dtype
```
wence-, line 146: both branches identical?

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+    def with_types(self, arg_id_to_dtype, kernel, callables_table):
+        for i in range(len(arg_id_to_dtype)):
+            if arg_id_to_dtype.get(i) is None:
+                # the types provided aren't mature enough to specialize the
+                # callable
+                return (self.copy(arg_id_to_dtype=arg_id_to_dtype),
+                        callables_table)
+
+        mat_dtype = arg_id_to_dtype[0].numpy_dtype
+        name_in_target = "inverse"
```
wence-, line 114: Ignores `self.name_in_target`

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+
+        return (self.copy(name_in_target=name_in_target,
+                          arg_id_to_dtype={-1: NumpyType(mat_dtype, target=loopy.CTarget()), 0: NumpyType(mat_dtype, target=loopy.CTarget()), 1: NumpyType(numpy.int32, target=loopy.CTarget())}),
```
wence-, line 117: Does one need the target here? The fact that it's not available via `self` or arguments makes me think not.

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
     return None
 
 
+class INVCallable(loopy.ScalarCallable):
+    def __init__(self, name, arg_id_to_dtype=None,
```
wence-, line 95: Brief doc describing what loopy call this matches?

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+
+        yield("solve", "#include <petscsys.h>\n#include <petscblaslapack.h>\n"+code)
+        return
+
+
+def sol_fn_lookup(target, identifier):
+    if identifier == 'solve':
+        return SolveCallable(name='solve')
+    else:
+        return None
```
wence-, line 295: These two callables (they are quite related) share a lot of code. Can we refactor so that they actually do so, and just subclass the bits that are different?

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+            #define BUF_SIZE 30
+
+            static PetscBLASInt ipiv_buffer[BUF_SIZE];
+            static void solve(PetscScalar* restrict out, const PetscScalar* restrict A, PetscScalar* restrict B, PetscBLASInt N)
+            {
+                PetscBLASInt info;
+                PetscBLASInt *ipiv = N <= BUF_SIZE ? ipiv_buffer : malloc(N*sizeof(*ipiv));
+                memcpy(out,B,N*sizeof(PetscScalar));
+                PetscBLASInt NRHS;
+                NRHS=1;
```
wence-, line 276: `const PetscBLASInt NRHS = 1;`

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+        assert isinstance(target, CTarget)
+        code = """#include <string.h>
+            #include <stdio.h>
+            #include <stdlib.h>
+            #ifndef Solve_HPP
+            #define Solve_HPP
+            #define BUF_SIZE 30
+
+            static PetscBLASInt ipiv_buffer[BUF_SIZE];
+            static void solve(PetscScalar* restrict out, const PetscScalar* restrict A, PetscScalar* restrict B, PetscBLASInt N)
```
wence-, line 270: Prefer `__restrict__` over `restrict`.

In both cases, I think `N` is compile-time known, so can we just inline it, rather than passing as a parameter?

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+                    fprintf(stderr,\"Gesv throws nonzero info.\");
+                if ( N > BUF_SIZE ) {
+                    free(ipiv);
+                }
+            }
+            #endif
+        """
+
+        yield("solve", "#include <petscsys.h>\n#include <petscblaslapack.h>\n"+code)
+        return
```
wence-, line 288: No need for bare return (and space between `yield` and `(`).

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+        c_parameters.insert(1, arg_c_parameters[0])  # A
+        c_parameters.insert(2, arg_c_parameters[1])  # B
+        c_parameters.insert(3, mat_descr_A.shape[1])  # n
+        return var(self.name_in_target)(*c_parameters), False
+
+    def generate_preambles(self, target):
+        assert isinstance(target, CTarget)
+        code = """#include <string.h>
+            #include <stdio.h>
+            #include <stdlib.h>
```
wence-, line 264: Rather than putting these includes here, shall we just whack them in the standard set of headers we require?

pyop2/codegen/rep2loopy.py
```diff
@@ -460,7 +670,7 @@ def generate(builder, wrapper_name=None):
     # register kernel
     kernel = builder.kernel
     headers = set(kernel._headers)
-    headers = headers | set(["#include <math.h>"])
+    headers = headers | set(["#include <math.h>"]) | set(["#include <stdint.h>"])
```
wence-, line 673: I think that `stdint.h` is now magically included by loopy, but maybe not. If not, probably write this as:

```python
headers |= {"#include <math.h>", "#include <stdint.h>"}
```

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+        return var(self.name_in_target)(*c_parameters), False
+
+    def generate_preambles(self, target):
+        assert isinstance(target, CTarget)
+        inverse_preamble = """
+            #include <string.h>
+            #include <stdio.h>
+            #include <stdlib.h>
+            #ifndef Inverse_HPP
+            #define Inverse_HPP
```
wence-, line 163: This is presumably to guard that this code is not emitted more than once. This seems ugly though, why does loopy not rationalise these generated preambles?

pyop2/codegen/rep2loopy.py
```diff
@@ -85,6 +91,210 @@ def petsc_function_lookup(target, identifier):
...
+            static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
+            static void inverse(PetscScalar* restrict Aout, const PetscScalar* restrict A, const PetscBLASInt N)
+            {
+                PetscBLASInt info;
+                PetscBLASInt *ipiv = N <= BUF_SIZE ? ipiv_buffer : malloc(N*sizeof(*ipiv));
+                PetscScalar *Awork = N <= BUF_SIZE ? work_buffer : malloc(N*N*sizeof(*Awork));
+                memcpy(Aout,A,N*N*sizeof(PetscScalar));
+                LAPACKgetrf_(&N,&N,Aout,&N,ipiv,&info);
+                if(info==0)
+                    LAPACKgetri_(&N,Aout,&N,ipiv,Awork,&N,&info);
```
wence-, line 176: Style:

```c
if (info == 0) {
   ...
}
```

Prefer braces even if there's a single statement.

pyop2/codegen/rep2loopy.py
wence-: If one does, prefer making a single target, rather than multiple ones of the same type. Also, please wrap the line a bit sooner:

```
{-1: ...,
 0: ..., 
 ...}
```

pyop2/codegen/rep2loopy.py
```diff
@@ -3,7 +3,11 @@
 
 import loopy
 from loopy.symbolic import SubArrayRef
-from loopy.types import OpaqueType
+from loopy.expression import dtype_to_type_context
+from pymbolic.mapper.stringifier import PREC_NONE
+from pymbolic import var
+from loopy.types import NumpyType, OpaqueType
+from loopy.target.c import CTarget
```
wence-, line 10: Imported name never used.

pyop2/codegen/rep2loopy.py
```diff
@@ -32,6 +36,8 @@
 from pyop2.codegen.representation import (PackInst, UnpackInst, KernelInst)
 from pytools import ImmutableRecord
 
+TARGET=loopy.target.c.CTarget()
```
wence-, line 39: Never used?

pyop2/codegen/rep2loopy.py
sv2518: I introduced all the constructs related to Target, when I came up with #580. I think I accidentally mangled them a bit, because I needed both for Tsslac at some point. I will drop the Targets here now, but might need them later.

pyop2/codegen/rep2loopy.py
sv2518: With the standard set of headers being defined in PyOP2 I assume? I don't think that will work until the inlining is done right. If we don't include the headers here, we will need to attach them to the slate kernel before the C device code (which is then passed to PyOP2) is generated from it, I believe. Would that be better? Then we might have headers in our kernels, which we do not always use however.

pyop2/codegen/rep2loopy.py
sv2518: Actually maybe they are. The first argument in the yield might be an identifier that prevents this.

pyop2/codegen/rep2loopy.py
sv2518: What? You mean around the +, no?

pyop2/codegen/rep2loopy.py
wence-: I think we put the headers the kernel advertises in the preamble (see rep2loopy.py line 480 or so)? But this is probably fine.

pyop2/codegen/rep2loopy.py
wence-: `yield ("inverse", ...)`, not `yield("inverse", ...)`

pyop2/codegen/rep2loopy.py
sv2518: N is inherited earlier from the size of the matrix A (line 235 + 257), so you do not need to define N in the according loopy call instruction. Why would it be better to inline?
And with inlining you mean something like `const PetscBLASInt N = sizeof((A))/sizeof((A)[0]);` inside the C function defined here, or do you mean generating a loopy statement for N in the slate kernel.

pyop2/codegen/rep2loopy.py
sv2518: ah

pyop2/codegen/rep2loopy.py
sv2518: Yes we do, but I think that one is only called on Pyop2 level and I need the preambles already on slate level for the workaround of compiling the slate kernel to C code (before passing back to pyop2). I have moved the preamble to the slate compiler now, so we don't need to change the callables after we fixed the kernel inlining.

pyop2/codegen/rep2loopy.py
wence-: So since this preamble is generated for a problem with a compile-time known N, we can have two versions:

1. no runtime memory allocation occurs and the code looks something like
```c

static void solve(PetscScalar* __restrict__ out, const PetscScalar * __restrict__ A, const PetscScalar * __restrict__ B)
{
   PetscBLASInt ipiv[literal_value_of_N];
   memcpy(out,B,N*sizeof(*out));
  const PetscBLASInt NRHS = 1;
  const PetscBLASInt N = literal_value_of_N;
  LAPACKgesv_(&N, &NRHS, A, &N, ipiv, out, &N, &info);
  ...
}
```

And another version where we have:

```
static void solve(PetscScalar* __restrict__ out, const PetscScalar * __restrict__ A, const PetscScalar * __restrict__ B)
{
   PetscBLASInt *ipiv = malloc(literal_value_of_N * sizeof(*ipiv));
   memcpy(out,B,N*sizeof(*out));
  const PetscBLASInt NRHS = 1;
  const PetscBLASInt N = literal_value_of_N;
  LAPACKgesv_(&N, &NRHS, A, &N, ipiv, out, &N, &info);
  ...
  free(ipiv);
}
```

sv2518: Where is literal_value_of_N coming from?

wence-: Ah, I had thought we make these things when `N` is readily available to us, but perhaps it is not. If so, ignore me.

sv2518:
Thanks for the review, Lawrence. I believe I have addressed all of your comments.

sv2518:
It is used in Tsslac, but there is no separate test for it in PyOP2. I am not quite sure how I would do that.

test/unit/test_callables.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+@pytest.fixture
+def zero_vec(s):
+    return op2.Dat(s ** (2, 1), [0.0, 0.0])
+
+
+@pytest.fixture
+def solve_mat(s):
+    d = op2.Dat(s ** (2, 2), [[2.0, 3.0], [-1.0, 2.0]])
```
sv2518, line 63: The "solve" test fails. The result gives a solution for a system matrix that is the transpose of this matrix. Did I do something wrong on PyOP level in the setup of the test case?

sv2518: Resolved. I replaced gesv with a getrs call such that we can specify to transpose. It was a row major vs column major issue.

wence-: Oh yeah, fortran.

PR #583, opened 2020-05-13
Complex updates

wence-:
This is the result of the complex sprint in PyOP2. For review, but the ordering requirements need the loopy changes to land first.


wence-:
PyOP2 master runs with loopy complex updates (so that's fine).

PR #587, opened 2020-06-09
wence/mixed matrix assembly

wence-:
Long-awaited, implements assembly into mixed matrices. The kernel should write to a single output array which is then split apart and inserted into the correct blocks.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -773,11 +773,6 @@ class Subset(ExtrudedSet):
         else:
             return self._superset.layers_array[self.indices, ...]
 
-    @cached_property
-    def _argtype(self):
-        """Ctypes argtype for this :class:`Subset`"""
-        return ctypes.c_voidp
-
 
 class SetPartition(object):
     def __init__(self, set, offset, size):
@@ -1410,11 +1405,6 @@ class Dat(DataCarrier, _EmptyDataMixin):
         """The scalar number of values for each member of the object. This is
         the product of the dim tuple."""
         return self.dataset.cdim
-
-    @cached_property
-    def _argtype(self):
-        """Ctypes argtype for this :class:`Dat`"""
-        return ctypes.c_voidp
 
     @property
     @collective
@@ -2329,11 +2319,6 @@ class Global(DataCarrier, _EmptyDataMixin):
         return _make_object('GlobalDataSet', self)
 
     @property
-    def _argtype(self):
-        """Ctypes argtype for this :class:`Global`"""
-        return ctypes.c_voidp
-
-    @property
     def shape(self):
         return self._dim
 
@@ -2551,11 +2536,6 @@ class Map(object):
     def __len__(self):
         """This is not a mixed type and therefore of length 1."""
         return 1
-
-    @cached_property
-    def _argtype(self):
-        """Ctypes argtype for this :class:`Map`"""
-        return ctypes.c_voidp
 
     @cached_property
     def split(self):
@@ -3160,12 +3140,7 @@ class Mat(DataCarrier):
     @cached_property
     def _argtypes_(self):
         """Ctypes argtype for this :class:`Mat`"""
-        return (ctypes.c_voidp, )
-
-    @cached_property
-    def _argtype(self):
-        """Ctypes argtype for this :class:`Mat`"""
-        return ctypes.c_voidp
+        return tuple(ctypes.c_voidp for _ in self)
 
     @cached_property
     def dims(self):
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3531,9 +3531,15 @@ class ParLoop(object):
         with self._parloop_event:
             orig_lgmaps = []
             for arg in self.args:
-                if arg._is_mat and arg.lgmaps is not None:
-                    orig_lgmaps.append(arg.data.handle.getLGMap())
-                    arg.data.handle.setLGMap(*arg.lgmaps)
+                if arg._is_mat:
+                    new_state = {INC: Mat.ADD_VALUES,
+                                 WRITE: Mat.INSERT_VALUES}[arg.access]
+                    for m in arg.data:
+                        m.change_assembly_state(new_state)
+                    arg.data.change_assembly_state(new_state)
+                    if arg.lgmaps is not None:
+                        orig_lgmaps.append(arg.data.handle.getLGMap())
+                        arg.data.handle.setLGMap(*arg.lgmaps)
             self.global_to_local_begin()
             iterset = self.iterset
             arglist = self.arglist
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -215,11 +215,11 @@ class Arg(object):
             return tuple(_make_object('Arg', d, m, self._access)
                          for d, m in zip(self.data, self._map))
         elif self._is_mixed_mat:
-            s = self.data.sparsity.shape
+            rows, cols = self.data.sparsity.shape
             mr, mc = self.map
             return tuple(_make_object('Arg', self.data[i, j], (mr.split[i], mc.split[j]),
                                       self._access)
-                         for j in range(s[1]) for i in range(s[0]))
+                         for i in range(rows) for j in range(cols))
         else:
             return (self,)
 
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -124,7 +124,7 @@ class Arg(object):
         :param map:  A :class:`Map` to access this :class:`Arg` or the default
                      if the identity map is to be used.
         :param access: An access descriptor of type :class:`Access`
-        :param lgmaps: For :class:`Mat` objects, a 2-tuple of local to
+        :param lgmaps: For :class:`Mat` objects, a tuple of 2-tuples of local to
             global maps used during assembly.
 
         Checks that:
@@ -148,6 +148,7 @@ class Arg(object):
         self.lgmaps = None
         if self._is_mat and lgmaps is not None:
             self.lgmaps = as_tuple(lgmaps)
+            assert len(self.lgmaps) == self.data.nblocks
         else:
             if lgmaps is not None:
                 raise ValueError("Local to global maps only for matrices")
@@ -3138,6 +3139,10 @@ class Mat(DataCarrier):
             "Abstract Mat base class doesn't know how to set values.")
 
     @cached_property
+    def nblocks(self):
+        return int(np.prod(self.sparsity.shape))
+
+    @cached_property
     def _argtypes_(self):
         """Ctypes argtype for this :class:`Mat`"""
         return tuple(ctypes.c_voidp for _ in self)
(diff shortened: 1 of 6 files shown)


pyop2/codegen/rep2loopy.py
```diff
@@ -311,9 +311,14 @@ def bounds(exprs):
...
 
+    # These are used in mixed mat unpacking
+    for op in instructions_by_type[PreUnpackInst]:
+        deps[op] |= frozenset(names[o] for o in instructions_by_type[KernelInst])
+
     # unpacking instructions depends on kernel instructions
     for op in instructions_by_type[UnpackInst]:
-        deps[op] |= frozenset(names[o] for o in instructions_by_type[KernelInst])
+        deps[op] |= frozenset(names[o] for o in itertools.chain(instructions_by_type[KernelInst],
+                                                                instructions_by_type[PreUnpackInst]))
```
sv2518, line 322: Maybe something like this?
```
inst_pack_map=OrderDict({KernelInst:[PackInst],
                         PreUnpackInst:[KernelInst],
                         UnpackInst:[KernelInst,PreUnpackInst]})
for inst in inst_pack_map.keys():
     for op in instructions_by_type[inst]:
          name_keys = itertools.chain(instructions_by_type[pack] for pack in inst_pack_map[inst])
          deps[op] |= frozenset(names[o] for o in name_keys)
```

pyop2/base.py
```diff
@@ -215,11 +216,11 @@ def split(self):
             return tuple(_make_object('Arg', d, m, self._access)
                          for d, m in zip(self.data, self._map))
         elif self._is_mixed_mat:
-            s = self.data.sparsity.shape
+            rows, cols = self.data.sparsity.shape
             mr, mc = self.map
             return tuple(_make_object('Arg', self.data[i, j], (mr.split[i], mc.split[j]),
                                       self._access)
-                         for j in range(s[1]) for i in range(s[0]))
+                         for i in range(rows) for j in range(cols))
```
sv2518, line 223: Is iterating over the rows first faster than over the columns?

pyop2/codegen/builder.py
```diff
@@ -442,6 +453,91 @@ def emit_unpack_instruction(self, *, loop_indices=None):
         yield call
 
 
+class MixedMatPack(Pack):
+
+    insertion_names = {False: "MatSetValuesBlockedLocal",
+                       True: "MatSetValuesLocal"}
+    """Function call name for inserting into the PETSc Mat. The keys
+       are whether or not maps are "unrolled" (addressing dofs) or
```
sv2518, line 461: What keys? The insertion_names?

pyop2/codegen/builder.py
```diff
@@ -442,6 +453,91 @@ def emit_unpack_instruction(self, *, loop_indices=None):
...
+       blocked (addressing nodes)."""
+
+    def __init__(self, packs, access, dtype, block_shape, interior_horizontal=False):
+        self.access = access
+        assert len(block_shape) == 2
```
sv2518, line 466: So we cannot assemble mixed vectors? Do we have something like `MixedVecPack`s? Or is the shape here something like `block_shape =(n,1) `? Just asking because in Slate it would be `block_shape = (n,)`, I think.

pyop2/codegen/builder.py
```diff
@@ -442,6 +453,91 @@ def emit_unpack_instruction(self, *, loop_indices=None):
...
+        if self.interior_horizontal:
+            shape = (2, )
+        else:
+            shape = (1, )
+        rshape = 0
+        cshape = 0
+        # Need to compute row and col shape based on individual pack shapes
+        for p in self.packs[:, 0]:
+            shape, _ = p.shapes
+            rshape += numpy.prod(shape, dtype=int)
```
sv2518, line 483: I do not know if it matters here, but maybe use IntType?

pyop2/codegen/builder.py
```diff
@@ -371,14 +373,23 @@ def pack(self, loop_indices=None):
             shape = (1, )
         rshape = shape + rmap.shape[1:] + (rdim, )
         cshape = shape + cmap.shape[1:] + (cdim, )
+        return (rshape, cshape)
```
sv2518, line 376: What information is encoded in rmap.shape[1:]? I am trying to understand why you multiply these values together in line 486.

pyop2/codegen/builder.py
```diff
@@ -442,6 +453,91 @@ def emit_unpack_instruction(self, *, loop_indices=None):
...
+        for row in self.packs:
+            coffset = 0
+            for p in row:
+                rshape, cshape = p.shapes
+                pack_ = p.pack(loop_indices=loop_indices, only_declare=True)
+                rindices = tuple(Index(e) for e in rshape)
+                cindices = tuple(Index(e) for e in cshape)
+                indices = MultiIndex(*rindices, *cindices)
+                lvalue = Indexed(pack_, indices)
+                rextents = [numpy.prod(rshape[i+1:], dtype=numpy.int32) for i in range(len(rshape))]
```
sv2518, line 519: Here maybe also IntType?

pyop2/codegen/builder.py
```diff
@@ -442,6 +453,91 @@ def emit_unpack_instruction(self, *, loop_indices=None):
...
+                                        Literal(IntType.type(0), casting=False))
+
+                flat_index = MultiIndex(Sum(flat_row_index, Literal(IntType.type(roffset), casting=False)),
+                                        Sum(flat_col_index, Literal(IntType.type(coffset), casting=False)))
+                rvalue = Indexed(pack, flat_index)
+                # Copy from local mixed element tensor into non-mixed
+                mixed_to_local.append(Accumulate(PreUnpackInst(), lvalue, rvalue))
+                # And into global matrix.
+                local_to_global.extend(p.emit_unpack_instruction(loop_indices=loop_indices))
+                coffset += numpy.prod(cshape, dtype=numpy.int32)
```
sv2518, line 535: Okay, so the way this works is we run over the mixed matrix, extract the non-mixed matrices and invoke the distribution of the local contributions into the global matrix on each of them separately? In theory, could Slate do mixed matrixes built of matrices that are themselves mixed? If yes, does the unpacking work for them as well?

pyop2/base.py
wence-: This is just so iterating over the `Arg` comes out in the same order as iterating over the `Mat` that is inside the `Arg`. Previously one was the transpose of the other.

pyop2/codegen/builder.py
wence-: `rmap.shape` is the shape of the map from the iteration set into the values at each entry of the set. So it's for example `(10, 3)` for a cell_node_map for P1 on 10 triangles. But really it's `(None, 3)` because the first entry is runtime known. So `rmap.shape[1:]` is the number of dofs that the kernel will see for this argument on the current set entity.

pyop2/codegen/builder.py
wence-: `False` and `True`. In fact, this is unnecessary since a `MixedMatPack` never inserts directly (instead inserting into the pieces).

pyop2/codegen/builder.py
wence-: There's a `MixedDatPack` a bit further up.

pyop2/codegen/builder.py
wence-: This doesn't matter. These numbers become literal shapes in the loopy representation (rather than encoding values that are passed in as arguments).

pyop2/codegen/builder.py
wence-: As above.

pyop2/codegen/builder.py
wence-: So right now PyOP2 doesn't represent the concept of multiple levels of nesting (e.g. a mixeddat is always just a flat bag of dats). As a consequence the blocks in the matrix are always non-mixed, so this code works.

pyop2/base.py
```diff
@@ -3556,9 +3536,18 @@ def compute(self):
...
+                                 WRITE: Mat.INSERT_VALUES}[arg.access]
+                    for m in arg.data:
+                        m.change_assembly_state(new_state)
+                    arg.data.change_assembly_state(new_state)
+                    if arg.lgmaps is not None:
+                        olgmaps = []
+                        for m, lgmaps in zip(arg.data, arg.lgmaps):
+                            olgmaps.append(m.handle.getLGMap())
+                            m.handle.setLGMap(*lgmaps)
+                        orig_lgmaps.append(olgmaps)
```
thomasgibson, line 3550: Can you explain a bit about what is happening here? I think I get what 3545--3550 are doing. Basically you're running over the "block matrices" and gathering the indirection map for local-to-global-block insertion. But why do we keep the original lgmaps? Wouldn't it be sufficient to just keep `olgmaps`?

pyop2/codegen/builder.py
```diff
@@ -442,6 +453,85 @@ def emit_unpack_instruction(self, *, loop_indices=None):
...
 
 
+class MixedMatPack(Pack):
+
+    def __init__(self, packs, access, dtype, block_shape, interior_horizontal=False):
+        self.access = access
+        assert len(block_shape) == 2
+        self.packs = numpy.asarray(packs).reshape(block_shape)
+        self.dtype = dtype
+        self.interior_horizontal = interior_horizontal
```
thomasgibson, line 463: What is `interior_horizontal` exactly? Is this part of the "extrusion hack" in PyOP2 currently? I'm also not quite sure why `shape` takes the form is does in lines 468---471 as a result of this boolean attribute.

pyop2/base.py
```diff
@@ -3575,7 +3564,8 @@ def compute(self):
             self.update_arg_data_state()
             for arg in reversed(self.args):
                 if arg._is_mat and arg.lgmaps is not None:
-                    arg.data.handle.setLGMap(*orig_lgmaps.pop())
+                    for m, lgmaps in zip(arg.data, orig_lgmaps.pop()):
+                        m.handle.setLGMap(*lgmaps)
```
thomasgibson, line 3568: `orig_lgmaps.pop()` is just `olgmaps` as defined above right? It looks like whatever was previously contained in `orig_lgmaps` is thrown out either way (in fact, it basically only ever contains `olgmaps`).

I guess basically what I'm saying is the variable `orig_lgmaps` appears to do nothing. You gather `olgmaps` above, add it to the empty `orig_lgmaps` list, then extract `olgmaps` via `orig_lgmaps.pop()`.

Perhaps I'm just being dense and not seeing the reason for `orig_lgmaps`

pyop2/base.py
wence-: So matrices come with local to global maps that map process local indices to global indices (for insertion). To apply boundary conditions we replace these lgmaps with "temporary" lgmaps that map the boundary rows/columns to negative global indices (and those insertions are then discarded).

These new lgmaps are parloop-specific (and so live on the arg, not the matrix). So here what we're doing is swapping out the lgmaps, then assembly happens (insertions etc...).

pyop2/base.py
wence-: Now we have to restore the swapped out lgmaps for the original ones (so we don't always discard entries in these rows/columns). So we walk do that here.

pyop2/codegen/builder.py
wence-: Yes. Although it turns out I don't actually need this property here.

pyop2/codegen/builder.py
```diff
@@ -442,6 +453,85 @@ def emit_unpack_instruction(self, *, loop_indices=None):
...
+
+    def pack(self, loop_indices=None):
+        if hasattr(self, "_pack"):
+            return self._pack
+        if self.interior_horizontal:
+            shape = (2, )
+        else:
+            shape = (1, )
```
wence-, line 471: Never used.

wence-: Fixed.

pyop2/codegen/builder.py
wence-: Removed.

pyop2/base.py
wence-: Added a comment.

wence-:
Ready for re-review I think.

PR #590, opened 2020-07-03
wence/feature/codegen updates

wence-:


wence-:
@sv2518 I think this branch (if you merge it in vectorisation) + wence/fix/interpolate-max in Firedrake fixes the test_real_space bug.

sv2518:
Yes, it does! Thanks!

PR #596, opened 2020-08-28
Composable Maps

ReubenHill:
Needed for https://github.com/firedrakeproject/firedrake/pull/1815

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -2614,7 +2614,7 @@ class Map(object):
 class ComposedMap(object):
     def __init__(self, *maps):
         """Representation of map[0](map[1](map[2], ...))"""
-        maps = tuple(maps)
+        maps = as_tuple(maps)
         self.maps = maps
         for m1, m2 in zip(maps[:-1], maps[1:]):
             if m1.iterset != m2.toset:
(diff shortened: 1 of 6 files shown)


wence-:
I never finished this because I didn't do the codegen. It looks like the current implementation is just a (complicated) wrapper around explicitly consing the composition. 

If you're going to do that (i think it's a bad idea), then why not just have a function that does that composition for you rather than all of this mess.

ReubenHill:
> explicitly consing the composition
What do you mean by this?

>If you're going to do that (i think it's a bad idea), then why not just have a function that does that composition for you rather than all of this mess.

I might just not be following you, but it's not clear to me yet quite how you would compose a double indirection map. Shoudl I be able to work it out from looking at this code?

wence-:
> I might just not be following you, but it's not clear to me yet quite how you would compose a double indirection map. Should I be able to work it out from looking at this code?

Suppose you have `A : x -> y` and `B : y -> z`, then `C = B o A : x -> z`. This is a composed map.

You represent `A` by a (dense) matrix of shape `dim(x) x arity(A)` (say 10). and `B` by something of `dim(y) x arity(B)` (say 5). Now `C = B o A` _can_ be represented by a dense matrix of `dim(x) x arity(A) x arity(B)`. Or, for a general composition of `k` maps, `C_k := o_{i=1}^k A_i : x -> z` something of `dim(x) x \prod_{i=1}^k arity(A_i)`.

My contention is that you don't want to explicitly cons the storage for C_k, since that requires something that scales with `\prod_i arity(A_i)` rather than `\sum_i arity(A_i)`. Instead, one should represent `C_k` symbolically, and teach the codegen engine how to materialise the map values one element of `x` at a time, as necessary.

As far as I can tell, reading the diff, this merge doesn't do that, it just makes a `ComposedMap` object that materialises the entire storage.

[Aside: It also introduces a bunch of other stuff that seems to be completely unrelated to composition of maps.]

So I think if you want composed maps, the best place to start is the first commit in this series, and teach the codegen engine how to handle composed maps (I did not manage that).

ReubenHill:
What do you mean by `cons` here?

wence-:
[Memory allocation](https://en.wikipedia.org/wiki/Cons#Use_in_conversation)

ReubenHill:
I do not currently have plans to implement this - closing.

PR #597, opened 2020-08-30
Outsource the linear algebra c functions 

sv2518:
Is there any reason not to outsource the linear algebra c functions (inverse, solve) into separate files and read them in as strings for the preamble of the respective loopy Callable?

That would make it a) easier for me to code them and b) the code in rep2loopy more readable, in particular with more c kernels coming for vectorisation with batched blas and vectorisation without batched blas.

### DIFF
--- a/pyop2/codegen/c/inverse.c
+++ b/pyop2/codegen/c/inverse.c
@@ -1,0 +1,29 @@
+#include <petscsys.h>
+#include <petscblaslapack.h>
+
+#ifndef PYOP2_WORK_ARRAYS
+#define PYOP2_WORK_ARRAYS
+#define BUF_SIZE 30
+static PetscBLASInt ipiv_buffer[BUF_SIZE];
+static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
+#endif
+
+static void inverse(PetscScalar* __restrict__ Aout, const PetscScalar* __restrict__ A, PetscBLASInt N)
+{
+    PetscBLASInt info;
+    PetscBLASInt *ipiv = N <= BUF_SIZE ? ipiv_buffer : malloc(N*sizeof(*ipiv));
+    PetscScalar *Awork = N <= BUF_SIZE ? work_buffer : malloc(N*N*sizeof(*Awork));
+    memcpy(Aout, A, N*N*sizeof(PetscScalar));
+    LAPACKgetrf_(&N, &N, Aout, &N, ipiv, &info);
+    if(info == 0){
+        LAPACKgetri_(&N, Aout, &N, ipiv, Awork, &N, &info);
+    }
+    if(info != 0){
+        fprintf(stderr, "Getri throws nonzero info.");
+        abort();
+    }
+    if ( N > BUF_SIZE ) {
+        free(Awork);
+        free(ipiv);
+    }
+}
--- a/pyop2/codegen/c/solve.c
+++ b/pyop2/codegen/c/solve.c
@@ -1,0 +1,33 @@
+#include <petscsys.h>
+#include <petscblaslapack.h>
+
+#ifndef PYOP2_WORK_ARRAYS
+#define PYOP2_WORK_ARRAYS
+#define BUF_SIZE 30
+static PetscBLASInt ipiv_buffer[BUF_SIZE];
+static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
+#endif
+
+static void solve(PetscScalar* __restrict__ out, const PetscScalar* __restrict__ A, const PetscScalar* __restrict__ B, PetscBLASInt N)
+{
+    PetscBLASInt info;
+    PetscBLASInt *ipiv = N <= BUF_SIZE ? ipiv_buffer : malloc(N*sizeof(*ipiv));
+    memcpy(out,B,N*sizeof(PetscScalar));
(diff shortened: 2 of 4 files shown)


pyop2/codegen/c/inverse.c
```diff
@@ -0,0 +1,24 @@
...
+    }
+    if(info != 0){
+        fprintf(stderr, "Getri throws nonzero info.");
+        abort();
+    }
+    if ( N > BUF_SIZE ) {
+        free(Awork);
+        free(ipiv);
+    }
+}
```
wence-, line 24: Missing newline at end of file.

pyop2/codegen/c/inverse.c
```diff
@@ -0,0 +1,24 @@
+#define Inverse_HPP
```
wence-, line 1: This define is unused I think. It was maybe used as an include guard at some point?

pyop2/codegen/rep2loopy.py
```diff
@@ -160,33 +160,10 @@ class INVCallable(LACallable):
...
-                    free(Awork);
-                    free(ipiv);
-                }
-            }
-        """
-        yield ("inverse", "#include <petscsys.h>\n#include <petscblaslapack.h>\n" + inverse_preamble)
+        import os
+        with open(os.path.dirname(__file__)+"/c/inverse.c", "r") as myfile:
+            inverse_preamble = myfile.read()
+            yield ("inverse", "#include <petscsys.h>\n#include <petscblaslapack.h>\n" + inverse_preamble)
```
wence-, line 166: `#include` the petsc headers in the source, rather than pasting them on here?

pyop2/codegen/rep2loopy.py
```diff
@@ -160,33 +160,10 @@ class INVCallable(LACallable):
...
-                }
-                if ( N > BUF_SIZE ) {
-                    free(Awork);
-                    free(ipiv);
-                }
-            }
-        """
-        yield ("inverse", "#include <petscsys.h>\n#include <petscblaslapack.h>\n" + inverse_preamble)
+        import os
+        with open(os.path.dirname(__file__)+"/c/inverse.c", "r") as myfile:
```
wence-, line 164: Can you open and read these files once (on import on in a prepopulation phase) rather than every time. Additionally, we probably want to read on only one rank on any given communicator (this might require refactoring so that the codegen builder knows about communicators).

wence- (changes requested):
I don't think the empty `codegen/c/__init__.py` file is necessary, but you do need to add all these files to the `package_data` list in `setup.py`.

Some other comments to do with when we read the files and so forth.

pyop2/codegen/rep2loopy.py
sv2518: > Can you open and read these files once (on import on in a prepopulation phase) rather than every time. 
Like that https://github.com/OP2/PyOP2/pull/597/commits/da1de082ebbd25a3598d8bed59ed6bbebd015cdf ?

sv2518: > Additionally, we probably want to read on only one rank on any given communicator (this might require refactoring so that the codegen builder knows about communicators).

I think I will need help with that/more information how exactly to achieve that. Is there are a similar example somewhere?

wence-: If you have a communicator, then you can do:

```
if comm.rank == 0:
    with open(...) as f:
        source = f.read()
else:
    source = None
source = comm.bcast(source, root=0)
```

sv2518: Okay, and what do you mean with refactoring?

Can we not just do the following?
```
from pyop2.mpi import COMM_WORLD
if COMM_WORLD.rank == 0:
    with open(...) as f:
        source = f.read()
else:
    source = None
source = COMM_WORLD.bcast(source, root=0)
```

pyop2/codegen/c/solve.c
```diff
@@ -0,0 +1,29 @@
+#include <petscsys.h>
+#include <petscblaslapack.h>
+#define BUF_SIZE 30
+
+static PetscBLASInt ipiv_buffer[BUF_SIZE];
+static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
```
wence-, line 6: If we include both solve and inverse, then these variables are declared twice, which I think will not work.

Similarly, the BUF_SIZE will be `#define`d twice. 

the `BUF_SIZE` is fine because you can `#undef` it at the end of the function.

For the work arrays we can either name them uniquely for these two files, or wrap them both in include guards. In the latter case we should also define BUF_SIZE only once (so that we don't accidentally change one but not the other).

```c
#ifndef PYOP2_WORK_ARRAYS
#define PYOP2_WORK_ARRAYS
#define BUF_SIZE 30
static PetscBLASInt ipiv_buffer[BUF_SIZE];
#endif
```

setup.py
```diff
@@ -149,7 +149,7 @@ def run(self):
       test_requires=test_requires,
       packages=['pyop2', 'pyop2.codegen'],
       package_data={
-          'pyop2': ['assets/*', '*.h', '*.pxd', '*.pyx']},
+          'pyop2': ['assets/*', '*.h', '*.pxd', '*.pyx', '*.c']},
```
wence-, line 152: Does this pick up the C files, I think it needs to be `'codegen/c/*.c'` (If I read https://docs.python.org/3/distutils/setupscript.html#installing-package-data correctly).

PR #604, opened 2020-11-27
Fix API documentation build problems

connorjward:
Due to Sphinx having to import all of the files that it intends to document, it requires a working PETSc install in order to not crash. The easiest way to achieve this is to use the Firedrake Docker container (the full version is used rather than the vanilla one because it includes `--documentation-dependencies`). 

Also, since PyOP2 contains some Cython that needs compiling before imports are allowed, the easiest thing to do is to use the preinstalled PyOP2 that is part of the Firedrake stack (hence the `git pull` command).


wence-:
Thanks.

I guess the other option is to migrate the build-testing (currently done on travis) to actions (then build the docs at the end of that)

WDYT?

connorjward:
I think that's a good idea. Having both actions and travis could get confusing and it will avoid repeated builds. Also I really dislike my solution to this!

I'll get on that.

connorjward:
I forgot that pushing to the branch would register here so that 'Add CI action' commit is garbage along with all the force pushes. Anyway. I think it should now hopefully work. It certainly seems to on my fork.

I added two workflows, one for building the docs and one for testing, because they are triggered on totally different actions (push to master and PR).

dham (approved):

PR #605, opened 2020-12-14
Replace deprecated new_hash

connorjward:


ksagiyam:
This change itself looks good to me, but tests are failing (probably because it uses current `firedrake-project/loopy`).

connorjward:
Yeah that's why. What I've done shouldn't affect anything at all. `new_hash` was just calling `sha256` under the hood anyway.

inducer:
Oh yikes. Sorry! I had new idea you guys were relying on that. :grimacing:

wence-:
Neither did we, I think

inducer:
FYI https://github.com/inducer/pytools/pull/60

PR #608, opened 2021-01-22
Test code on merge with master

### LINKED ISSUE #607: CI doesn't test merge commits to master any more
Since #604, we test the merge commit in the PR, but don't actually test the final merge. Since our merges are not always fast-forwards, this might result in missed breakages.

I think we should also build-test on the push to trunk, which I guess is a case of changing the `on` specifier in ci.yml to be the union of the pull_request and push+master things.

@connorjward is this easy to do?

connorjward:
Resolves #607

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,14 @@
 name: CI
 
-on: pull_request
+# Trigger the workflow on push or pull request,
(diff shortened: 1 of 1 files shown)


wence- (approved):

wence-:
Thanks!

PR #609, opened 2021-04-21
Fix disjoint arguments to C-string kernels

connorjward:
I just want to make sure that I've interpreted your suggestions correctly. Tests are now passing.

@wence- please double check.

### DIFF
--- a/pyop2/codegen/rep2loopy.py
+++ b/pyop2/codegen/rep2loopy.py
@@ -204,13 +204,12 @@ class PyOP2KernelCallable(loopy.ScalarCallable):
     """Handles PyOP2 Kernel passed in as a string
     """
 
-    fields = set(["name", "arguments", "access", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target"])
-    init_arg_names = ("name", "arguments", "access", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target")
+    fields = set(["name", "parameters", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target"])
+    init_arg_names = ("name", "parameters", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target")
 
-    def __init__(self, name, arguments, access, arg_id_to_dtype=None, arg_id_to_descr=None, name_in_target=None):
+    def __init__(self, name, parameters, arg_id_to_dtype=None, arg_id_to_descr=None, name_in_target=None):
         super(PyOP2KernelCallable, self).__init__(name, arg_id_to_dtype, arg_id_to_descr, name_in_target)
-        self.arguments = arguments
-        self.access = access
+        self.parameters = parameters
 
     def with_types(self, arg_id_to_dtype, callables_table):
         new_arg_id_to_dtype = arg_id_to_dtype.copy()
@@ -237,8 +236,7 @@ class PyOP2KernelCallable(loopy.ScalarCallable):
 
     def emit_call_insn(self, insn, target, expression_to_code_mapper):
         # reorder arguments, e.g. a,c = f(b,d) to f(a,b,c,d)
-        parameters = self.arguments[insn.id]
-        par_dtypes = tuple(expression_to_code_mapper.infer_type(p) for p in parameters)
+        par_dtypes = tuple(expression_to_code_mapper.infer_type(p) for p in self.parameters)
 
         from loopy.expression import dtype_to_type_context
         from pymbolic.mapper.stringifier import PREC_NONE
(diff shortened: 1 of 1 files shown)


pyop2/codegen/rep2loopy.py
```diff
@@ -655,12 +649,21 @@ def statement_functioncall(expr, context):
...
                                  predicates=predicates,
                                  id=id,
                                  depends_on=depends_on, depends_on_is_final=True)
 
+    # We cannot reconstruct the correct calling convention for C-string kernels
+    # without providing some additional context about the argument ordering.
+    # This is processed inside the ``emit_call_insn`` method of
+    # :class:`.PyOP2KernelCallable`.
+    if not hasattr(context, "function_args"):
+        context.function_args = {}
```
wence-, line 663: Just set this up in the `WrapperBuilder` constructor?

pyop2/codegen/rep2loopy.py
```diff
@@ -655,12 +649,21 @@ def statement_functioncall(expr, context):
...
+    context.function_args[id] = function_args
```
wence-, line 664: Do we guarantee that the `id` is unique? I guess so, is a `CallInstruction` hashable, that would make me less nervous.

pyop2/codegen/rep2loopy.py
```diff
@@ -204,11 +204,12 @@ class PyOP2KernelCallable(loopy.ScalarCallable):
...
-    fields = set(["name", "access", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target"])
-    init_arg_names = ("name", "access", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target")
+    fields = set(["name", "arguments", "access", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target"])
+    init_arg_names = ("name", "arguments", "access", "arg_id_to_dtype", "arg_id_to_descr", "name_in_target")
 
-    def __init__(self, name, access, arg_id_to_dtype=None, arg_id_to_descr=None, name_in_target=None):
+    def __init__(self, name, arguments, access, arg_id_to_dtype=None, arg_id_to_descr=None, name_in_target=None):
         super(PyOP2KernelCallable, self).__init__(name, arg_id_to_dtype, arg_id_to_descr, name_in_target)
+        self.arguments = arguments
         self.access = access
```
wence-, line 213: I think that access is no longer used so we could remove it?

pyop2/codegen/rep2loopy.py
```diff
@@ -562,7 +551,9 @@ def generate(builder, wrapper_name=None):
         wrapper = loopy.register_callable(
             wrapper,
             kernel.name,
-            PyOP2KernelCallable(name=kernel.name, access=tuple(builder.argument_accesses)))
+            PyOP2KernelCallable(name=kernel.name,
+                                arguments=context.function_args,
```
wence-, line 555: Ah, at this point we don't have the instruction available, so we pass the whole dict. Is the name a good one?

pyop2/codegen/rep2loopy.py
connorjward: Now set up where `context` is defined.

pyop2/codegen/rep2loopy.py
connorjward: `CallInstruction` objects aren't hashable. I've used the `name` attribute instead now.

pyop2/codegen/rep2loopy.py
connorjward: If you mean "can we key the dict using `kernel.name`? then the answer is yes and I've done that.

pyop2/codegen/rep2loopy.py
sv2518: Maybe it is possible to use the function here https://github.com/firedrakeproject/loopy/blob/94f191ce2545de5c959b19844d367c129473e80d/loopy/kernel/instruction.py#L441?

connorjward: Is the kernel `name` attribute not unique?

wence-: It must be, yes.

pyop2/codegen/rep2loopy.py
wence-: I meant is `function_args` a good name for this dict?

connorjward: `kernel_parameters`?

wence-: Better, I think.

connorjward: Done.

wence- (approved):
Looks right, thanks.

PR #610, opened 2021-04-21
Pin decorator

dham:
Decorator version 5 changes the API in ways that break the
argument type checking in PyOP2. We're therefore pinning on the
version that works.


connorjward (approved):
This fixed it for me.

wence-:
Can we open an issue to track this? Long term I would prefer not to be pinned to an old version.

wence-:
Did that in #611.

PR #613, opened 2021-04-23
(DO NOT MERGE) Fix conflicting code generation results

connorjward:
This change fixes (at least for me) the `CompilationError "Generated code differs across ranks"` error that is getting thrown by `tests/multigrid/test_poisson_gmg_extruded_serendipity.py`.

I have absolutely no idea why this works. We shouldn't have modified anything to do with the communicators.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -3415,7 +3415,7 @@ class JITModule(Cached):
     def _cache_key(cls, kernel, iterset, *args, **kwargs):
         counter = itertools.count()
         seen = defaultdict(lambda: next(counter))
-        key = ((id(dup_comm(iterset.comm)), ) + kernel._wrapper_cache_key_ + iterset._wrapper_cache_key_
+        key = (() + kernel._wrapper_cache_key_ + iterset._wrapper_cache_key_
                + (iterset._extruded, (iterset._extruded and iterset.constant_layers), isinstance(iterset, Subset)))
 
         for arg in args:


pyop2/base.py
```diff
@@ -3415,7 +3415,7 @@ class JITModule(Cached):
     def _cache_key(cls, kernel, iterset, *args, **kwargs):
         counter = itertools.count()
         seen = defaultdict(lambda: next(counter))
-        key = ((id(dup_comm(iterset.comm)), ) + kernel._wrapper_cache_key_ + iterset._wrapper_cache_key_
+        key = (() + kernel._wrapper_cache_key_ + iterset._wrapper_cache_key_
```
kaushikcfd, line 3418: At least for compilation this is the first entry is ignored: https://github.com/OP2/PyOP2/blob/loopy-sprint/pyop2/compilation.py#L239

wence-: Kaushik is right. This change is not the right thing to do. Check git blame for why the communicator is part of the key.

wence- (changes requested):

pyop2/base.py
wence-: https://github.com/OP2/PyOP2/commit/2e1a4b7a607e56602dc9aee19d06048e9d2b14ed

connorjward: I totally agree with both of you, this 'fix' is awful. The thing is is that somehow this change got the test passing for me. I know not how.

I created this PR so others could see what progress I'd made and discuss it. I should have made clear that I didn't want it merged.

connorjward:
I have identified my idiocy. It is in fact `kernel._wrapper_cache_key_` that is the culprit. As @kaushikcfd [pointed out](https://github.com/OP2/PyOP2/blob/loopy-sprint/pyop2/compilation.py#L239) the first element of the cache key is ignored for compilation. Hence when I replace the `id(dup_comm(comm))` with an empty tuple the first element of the cache key that gets removed is in fact `kernel._wrapper_cache_key_`.

PR #616, opened 2021-05-10
Trigger github actions rerun.

sv2518:
I think master is currently failing in the installation of mpi4py. One can see that in Connor's and the loopy sprint branch. Opening a PR to rerun the actions on master to double-check if what I am saying is indeed the case.


connorjward:
It's the petsc4py install that's failing not mpi4py.

wence-:
`Cloning https://github.com/firedrakeproject/petsc4py.git (to revision firedrake)`

The actions workflow needs to be updated to grab petsc4py from the petsc source tree.

I note also that currently we're installing petsc from pip, rather than the firedrake petsc clone. So plausibly that wants reworked too.

connorjward:
I've agreed to take this on. Do we need any special configure options for PETSc when we build it or is the default (plus petsc4py) sufficient?

wence-:
Pretty sure default is sufficient (we're just doing `pip install petsc` right now as far as I recall). Probably `--with-shared-libraries=1 --with-debugging=1` is a good choice.

sv2518:
A fix got merged in f1daf214c1eedd86a37aaada2565c4d442c8ebbf.

PR #617, opened 2021-05-10
Make PETSc and petsc4py for CI use Firedrake not pip

connorjward:
This worked on my fork. I've also included some fixes from the `loopy-sprint` branch as these were not Loopy specific and would make our tests fail.

See also: #616


wence- (approved):

wence-:
Can you make the same changes in the gh-pages workflow? We should probably refactor so they share setup?

PR #619, opened 2021-05-11
Remove GH pages workflow and merge into CI one

connorjward:
This will reduce both overhead and boilerplate.


.github/workflows/ci.yml
```diff
@@ -79,3 +78,19 @@ jobs:
...
         run: pytest test -v --tb=native
+
+      - name: Build documentation
+        if: ${{ github.event_name == 'push' }}
+        shell: bash
+        working-directory: PyOP2
+        run: make -C doc/sphinx html
+
+      - name: Upload to GitHub Pages
+        if: ${{ github.event_name == 'push' }}
```
wence-, line 89: Can we guard this on pushes to master only as well?

.github/workflows/ci.yml
```diff
@@ -79,3 +78,19 @@ jobs:
         shell: bash
         working-directory: PyOP2
         run: pytest test -v --tb=native
+
+      - name: Build documentation
+        if: ${{ github.event_name == 'push' }}
```
wence-, line 83: This one wants to always run, to ensure that the docs actually build!

wence- (changes requested):

wence-:
I did kind of the same thing in #618. I also made the petsc build a bit faster and flipped to mpich, do you want to check that one? I don't mind which we merge but we should get all the changes.

.github/workflows/ci.yml
connorjward: This workflow only gets triggered on either pull requests or pushes to master.

wence-:
OK.

PR #621, opened 2021-05-14
Permuted maps

wence-:
Add an API for specifying `map o permutation` for some constant permutation of the map order. This is convenient in some cases where one wants to write local kernels that need a different ordering to the FIAT one (say).

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -2623,6 +2623,41 @@ class Map(object):
         return self == o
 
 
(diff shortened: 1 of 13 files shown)


wence-:
@ReubenHill I requested a review from you in case you want to see how to do codegen for map generation (in case you want to do map composition like this rather than up front.

dham (approved):

PR #624, opened 2021-06-01
Purge parloop data references

connorjward:
This PR aims to redesign the parloop API to fix the data reference issues it has that prevent straightforward caching.

The basic idea is to make a distinction between information necessary for parloop *generation* and that for its *execution*. The arguments to the parloop constructor will be modified and the execution arguments will be passed in when calling `parloop.compute()`.

The exact API we present is definitely up for debate but I'm initially focused on writing something that works.


connorjward:
OK so this now works and all of the tests pass (apart from some to do with exceptions and caching). The code is very messy so I don't recommend a review yet.

I want to discuss what the API should look like. Getting this right now will save me a TON of time modifying every test that uses parloops.

At present in this PR the way a parloop is created and run is as follows:

```python
loop = op2.ParLoop(kernel, [dat0(op2.RW), dat1(op2.READ, map0), ...], **kwargs)
loop(iterset, dat0, dat1, map0)
```
and `ParLoop.__init__` has the signature:

```python
    def __init__(self, kernel, arguments, *,
                 extruded=False,
                 constant_layers=False,
                 subset=False,
                 iteration_region=None,
                 pass_layer_arg=False):
```

The key changes here are:

- We only instantiate the parloop with the minimum amount of information necessary for the code generation. It doesn't reference any data.
- The metadata for `iterset` are passed in as `kwargs` (e.g. `extruded`).
- `iterset` and the data structures are then passed in when the parloop is executed.
- The parloop is executed with `__call__` not `compute`. I made this change because it seemed neater.

I think the main considerations about the API are:

- How do we create the `Args`? I'm not a fan of the call syntax especially since the resulting object is totally unrelated to the input dat/mat/global.
- How we do call the parloop? At the moment we have to put the dats/mats/globals before the maps and get the order correct. Personally I quite like the idea of using tuples to associate data and maps. E.g.:

  ```python
  loop(iterset, dat0, (dat1, map0), (mat0, map1, map2), ...)
  ```

  Another idea I've had is we could give each `Arg` a `name` attribute and then we could do:

  ```python
  loop(iterset, name1=dat0, name3=(mat0, map1, map2), name2=(dat1, map0), ...)
  ```

There are also a few naming preferences that I have that aren't particularly essential:

- `Arg` -> `Argument` - I think having our own args can get quite confusing when combined with `*args`.
- `ParLoop` -> `Parloop` - I just think it's neater.

@wence-, @dham what are your thoughts?

wence-:
> OK so this now works and all of the tests pass (apart from some to do with exceptions and caching). The code is very messy so I don't recommend a review yet.
> 
> I want to discuss what the API should look like. Getting this right now will save me a TON of time modifying every test that uses parloops.

All of the tests use the functional interface `par_loop(...)`. This might not have to change too much.

> At present in this PR the way a parloop is created and run is as follows:
> 
> ```python
> loop = op2.ParLoop(kernel, [dat0(op2.RW), dat1(op2.READ, map0), ...], **kwargs)

What are the return values of these `__call__` methods. I guess `Argument` objects. But for code generation purposes what does the argument contain?

My vague idea had been that for codegen we need to know the local shape and the map shape (if not `None`), the access descriptor, and the packer class/object type.

I think that we could perhaps take this opportunity to unpick another of the big design flaws and move the access descriptors onto the `Kernel` object.

> loop(iterset, dat0, dat1, map0)

This ordering is because that's what the generated code expects, I guess.

> ```
> 
> and `ParLoop.__init__` has the signature:
> 
> ```python
>     def __init__(self, kernel, arguments, *,
>                  extruded=False,
>                  constant_layers=False,
>                  subset=False,
>                  iteration_region=None,
>                  pass_layer_arg=False):
> ```
> 
> The key changes here are:
> 
> * We only instantiate the parloop with the minimum amount of information necessary for the code generation. It doesn't reference any data.
> * The metadata for `iterset` are passed in as `kwargs` (e.g. `extruded`).
> * `iterset` and the data structures are then passed in when the parloop is executed.
> * The parloop is executed with `__call__` not `compute`. I made this change because it seemed neater.
> 
> I think the main considerations about the API are:
> 
> * How do we create the `Args`? I'm not a fan of the call syntax especially since the resulting object is totally unrelated to the input dat/mat/global.

It's not unrelated (if you don't have the Dat you don't know what argument metadata to make).

If we have access descriptors on the kernel, we could have an explicit `Dat.codegen_data(map_=None)`

> * How we do call the parloop? At the moment we have to put the dats/mats/globals before the maps and get the order correct. Personally I quite like the idea of using tuples to associate data and maps. E.g.:
>   ```python
>   loop(iterset, dat0, (dat1, map0), (mat0, map1, map2), ...)
>   ```

And then here we could use `Dat.parloop_data(map_=None)`

Names to be bikeshedded, so you would have:

```
loop = ParLoop(kernel, dat0.codegen_data(), dat1.codegen_data(map_=map0), ...)
loop(iterset, dat0.parloop_data(), dat1.parloop_data(map_=map0), ...)
```

>   Another idea I've had is we could give each `Arg` a `name` attribute and then we could do:
>   ```python
>   loop(iterset, name1=dat0, name3=(mat0, map1, map2), name2=(dat1, map0), ...)
>   ```
> 
> There are also a few naming preferences that I have that aren't particularly essential:
> 
> * `Arg` -> `Argument` - I think having our own args can get quite confusing when combined with `*args`.

This is fine.

> * `ParLoop` -> `Parloop` - I just think it's neater.

I'm less convinced by this, but not that fussed.

pyop2/arg.py
```diff
@@ -0,0 +1,194 @@
...
+    def is_reduction(self):
+        return self._access in {INC, MIN, MAX}
+
+
+class MatArg(Arg):
+
+    def __init__(self, access, dtype, unroll_map, dims, lgmaps, map_shapes, map_offsets):
+        super().__init__(access, dtype)
+        if lgmaps is not None:
+            lgmaps = as_tuple(lgmaps)
```
wence-, line 137: These are heavy data and so shouldn't live here I think.

connorjward: Didn't know that. Thanks.

connorjward:
> All of the tests use the functional interface par_loop(...). This might not have to change too much.

I've removed that interface because args don't reference data any more so it can't work. We could possibly find a way to get it to work but we'd need a new object to tie together data and symbolics.

> What are the return values of these `__call__` methods. I guess Argument objects. But for code generation purposes what does the argument contain?

Yes it is an `Arg` (this hasn't been changed from the original API). I've subclassed `Arg` so its contents can vary but it's things like datatype and access descriptor (see `arg.py`).

> I think that we could perhaps take this opportunity to unpick another of the big design flaws and move the access descriptors onto the Kernel object.

Can you elaborate? What are the issues with the current design? Also is this a TSFC level change or PyOP2?

> It's not unrelated (if you don't have the Dat you don't know what argument metadata to make).

Yes but the argument is not specific to that dat which I think the call syntax implies.

> If we have access descriptors on the kernel, we could have an explicit Dat.codegen_data(map_=None)

I think I'd prefer a `functools.singledispatch` `make_argument` function or similar as it more strongly suggests that we are not tied to a specific dat, but this would work fine.

> And then here we could use Dat.parloop_data(map_=None)

I like the idea of not passing the whole dat/mat/global in. That was another thing I didn't really like about my current solution.

wence-:
> > All of the tests use the functional interface par_loop(...). This might not have to change too much.
> 
> I've removed that interface because args don't reference data any more so it can't work. We could possibly find a way to get it to work but we'd need a new object to tie together data and symbolics.
> 
> > What are the return values of these `__call__` methods. I guess Argument objects. But for code generation purposes what does the argument contain?
> 
> Yes it is an `Arg` (this hasn't been changed from the original API). I've subclassed `Arg` so its contents can vary but it's things like datatype and access descriptor (see `arg.py`).
> 
> > I think that we could perhaps take this opportunity to unpick another of the big design flaws and move the access descriptors onto the Kernel object.
> 
> Can you elaborate? What are the issues with the current design? Also is this a TSFC level change or PyOP2?

The Kernel should really advertise how its arguments are accessed (rather than the parloop). The change would be in whatever constructs the `op2.Kernel` object and also in PyOP2.

So you would say:

```python
k = Kernel(some_code, arglist=[INC, READ, READ, WRITE, MAX, ...])
```

That way you can also validate when someone calls the kernel that the number of arguments is correct!


> > It's not unrelated (if you don't have the Dat you don't know what argument metadata to make).
> 
> Yes but the argument is not specific to that dat which I think the call syntax implies.

OK, sure.
 
> > If we have access descriptors on the kernel, we could have an explicit Dat.codegen_data(map_=None)
> 
> I think I'd prefer a `functools.singledispatch` `make_argument` function or similar as it more strongly suggests that we are not tied to a specific dat, but this would work fine.

I guess so. You have to pass the `dat` in though, I suppose (or else unpick it in the caller).

connorjward:
> The Kernel should really advertise how its arguments are accessed (rather than the parloop). The change would be in whatever constructs the op2.Kernel object and also in PyOP2.
>
> So you would say:
>
> `k = Kernel(some_code, arglist=[INC, READ, READ, WRITE, MAX, ...])`
>
> That way you can also validate when someone calls the kernel that the number of arguments is correct!

Is there any other information that would be better to tie to the kernel that the arguments? In particular I'm thinking of a mixed kernel like:

```c
static void inc(double *d, double *x) {
  d[0] += x[0]; d[1] += x[0];
}
```

Just from the kernel we can see that `d` is mixed. Should that be stored on the kernel too?

connorjward:
Also datatype. The kernel already knows that.

connorjward:
@dham has had a good idea about the new interface: rather than passing in `dim`, `arity` and other parameters required for code generation when we create the parloop, we can instead determine these from the UFL input to TSFC.

This significantly reduces the amount of prep work that needs to be done to execute a parloop since all of the argument metadata is encapsulated by the TSFC kernel.

Regarding how this would look in the code, I think we should have the following classes:

- `LocalKernel`
  This would effectively be the C code/Loopy kernel, dtype and intent.
- `GlobalKernel`
  This would consist of a `LocalKernel` plus argument metadata such as `dim` and `arity`. One of these could be generated from a TSFC kernel with `GlobalKernel.from_tsfc` say.

A parloop could then just be a function instead of a class as all of the code generation would be happening at the level of the `GlobalKernel`.

As an aside, I think that distinguishing 'local' and 'global' kernels in this way is beneficial as we tend to discuss kernels in a way where the one referred to is not clear.

wence-:
> @dham has had a good idea about the new interface: rather than passing in `dim`, `arity` and other parameters required for code generation when we create the parloop, we can instead determine these from the UFL input to TSFC.
> 
> This significantly reduces the amount of prep work that needs to be done to execute a parloop since all of the argument metadata is encapsulated by the TSFC kernel.
> 
> Regarding how this would look in the code, I think we should have the following classes:
> 
> * `LocalKernel`
>   This would effectively be the C code/Loopy kernel, dtype and intent.
> * `GlobalKernel`
>   This would consist of a `LocalKernel` plus argument metadata such as `dim` and `arity`. One of these could be generated from a TSFC kernel with `GlobalKernel.from_tsfc` say.
> 
> A parloop could then just be a function instead of a class as all of the code generation would be happening at the level of the `GlobalKernel`.
> 
> As an aside, I think that distinguishing 'local' and 'global' kernels in this way is beneficial as we tend to discuss kernels in a way where the one referred to is not clear.

The only wrinkle I can see here is that a `LocalKernel` can technically have multiple different `GlobalKernel`s that are all equally valid.

For example (although this isn't something that is done right now), given a kernel which increments a single dat:
```
void incr(double *A)
{
    for (i = 0; i < 3; i++) A[i] += 1;
}
```
These two parloops are both valid:

```
iterset = Set(...)
dset = Set(...)
dat1 = Dat(DataSet(dset, dim=1)))
dat2 = Dat(DataSet(iterset, dim=3)))
map1 = Map(iterset, dset, arity=3)

par_loop(k, iterset, dat1(INC, map1))
par_loop(k, iterset, dat2(INC))
```

In other words, the local kernel only cares about the shape of the staged data, and multiple different parloops can end up with the same staged data.

I guess you envisage something like:


```
k = LocalKernel(...)
g = GlobalKernel(k, metadata_about_shapes) # everything needed to generate code

parloop(g, iterset, (dat1, map1), (dat2, map2), ...)
```
?

connorjward:
I'm not sure that I fully understand your point. It is already the case that a `LocalKernel` (currently just called `Kernel`) has this ambiguity. For us to code generate we need to know extra information about the structure of the data hence us needing `Args` in the past and now a `GlobalKernel`.

> I guess you envisage something like:
>
> ```
> k = LocalKernel(...)
> g = GlobalKernel(k, metadata_about_shapes) # everything needed to generate code
>
> parloop(g, iterset, (dat1, map1), (dat2, map2), ...)
> ```

Yes that's exactly right. That would be the interface for the general case. For the case when the kernel originates from TSFC this could be abbreviated to:

```
k = ...  # some kernel generated by TSFC
g = GlobalKernel.from_tsfc(k)  # note that we don't have to pass any metadata here
parloop(g, iterset, ...)
```

dham:
> I'm not sure that I fully understand your point. It is already the case that a `LocalKernel` (currently just called `Kernel`) has this ambiguity. For us to code generate we need to know extra information about the structure of the data hence us needing `Args` in the past and now a `GlobalKernel`.
> 
> > I guess you envisage something like:
> > ```
> > k = LocalKernel(...)
> > g = GlobalKernel(k, metadata_about_shapes) # everything needed to generate code
> > 
> > parloop(g, iterset, (dat1, map1), (dat2, map2), ...)
> > ```
> 
> Yes that's exactly right. That would be the interface for the general case. For the case when the kernel originates from TSFC this could be abbreviated to:
> 
> ```
> k = ...  # some kernel generated by TSFC
> g = GlobalKernel.from_tsfc(k)  # note that we don't have to pass any metadata here
> parloop(g, iterset, ...)
> ```

... or we just change TSFC to return GlobalKernel objects.

connorjward:
I think that a comment clarifying how I have structured/will structure things would be useful:

- We will have `LocalKernel` and `WrapperKernel` classes encapsulating the local and global assembly kernels (I thought `GlobalKernel` was an issue because of `Global` objects being entirely different).
- These respectively accept lists of `LocalKernelArg` and `WrapperKernelArg` objects that provide appropriate metadata about the arguments. The `LocalKernelArg` stores dtype and intent (`READ`, `WRITE`, etc), and the `WrapperKernelArg` stores `dim`, `arity` and `offset`.
- If we make the `WrapperKernel` the 'heavy' object for codegen we can make a parloop just be a function. It would have signature: `parloop(wrapper_kernel, iterset, parloop_args)`; where `parloop_args` are `ParloopArg` objects and these contain the data and any maps that might be needed.

It's a bit convoluted as we require 3 different argument types (plus the [ones I have added to TSFC](https://github.com/firedrakeproject/tsfc/pull/257)), but I think it achieves the separation of information that we want. In Firedrake we can actually have very concise code if we write functions that convert between TSFC argument types and PyOP2 ones (e.g. [here](https://github.com/firedrakeproject/firedrake/pull/2172/files#diff-a17f9f7d347d1e36d5b04e849d399f23cfbfdb41a93d88ebad3cf4b4b0638ffcR750)).

sv2518:
Slate Kernels are inherited from TSFC Kernels. Is there any reason why Slate Kernels could make what's proposed here more complicated @wence- ?

connorjward:
I've also encountered another, entirely separate issue regarding map aliasing. In the existing PyOP2 code we examine the input maps to see if they are identical, and if they are we only pass the map through to the wrapper kernel **once** with `int32_t const  *__restrict__ mapN`. Since we now do the code generation *before* we pass in  the actual maps this check is no longer possible. To me the most straightforward fix is to remove the `__restrict__` attribute for the maps so we can safely alias them. Is this the right thing to do?

wence-:
> I've also encountered another, entirely separate issue regarding map aliasing. In the existing PyOP2 code we examine the input maps to see if they are identical, and if they are we only pass the map through to the wrapper kernel **once** with `int32_t const *__restrict__ mapN`. Since we now do the code generation _before_ we pass in the actual maps this check is no longer possible. To me the most straightforward fix is to remove the `__restrict__` attribute for the maps so we can safely alias them. Is this the right thing to do?

The UFL can tell you which maps are unique (because they correspond to the unique functionspaces in the form).

The reason to try and do it this way is to avoid packing the same map multiple times.

connorjward:
So I have been doing some thinking about this and have come to the unfortunate conclusion that I might be going about this the wrong way. Up to this point the operating assumption behind this work has been that *generating code using data-carrying objects is bad*. This is because we cannot cache generated code if it holds references to data without introducing a memory leak.

To implement this I have introduced a new stage into the parloop-execution pipeline where we explicitly generate the code. This so-called `WrapperKernel` is created with `WrapperKernelArg` objects that declare type information about the kernel's arguments (e.g. `dim`, `arity`, `offset`). These objects do not reference the data structures and so the `WrapperKernel` is safe to cache.

However, I think that enforcing the separation here is redundant and introduces a lot of unnecessary boilerplate. If we implement some sort of builder pattern so that we have

```python
@cachetools.cached(some_cache_key)
def make_wrapper_kernel(local_kernel, iterset, arguments):
    """Make a wrapper kernel.
    :arg local_kernel: op2.LocalKernel
    :arg iterset: the iteration set
    :arg arguments: list of op2.Arg objects (these carry data)
    :returns: an op2.WrapperKernel object (does not carry data)
    """
    return WrapperKernelBuilder(local_kernel, iterset, arguments).build()
```

then we can safely generate code with data-carrying objects while still having a cache-able output. We just have to make sure that the `Arg` objects are not part of the cache key - and they already have a `cache_key` property that could be used.

I think that this approach will achieve the original goal of this work (i.e. make caching safer, easier and 'less magic'). I just wanted to draw attention in this post to the fact that *generating code with data-carrying inputs is actually fine*, and so what I have currently implemented  is significant overkill.

wence-:
Part of the rationale for wanting to separate this information is that at present we _need_ data carrying objects to generate code, which somehow feels wrong given that code generation really only needs metadta.

pyop2/global_kernel.py
```diff
@@ -0,0 +1,387 @@
...
+    @mpi.collective
+    def __call__(self, comm, *args):
+        """Execute the compiled kernel.
+
+        :arg comm: Communicator the execution is collective over.
+        :*args: Arguments to pass to the compiled kernel.
+        """
+        # If the communicator changes then we cannot safely use the in-memory
+        # function cache.
+        key = id(mpi.dup_comm(comm))
```
wence-, line 289: This dup_comm is unnecessary, since the comm coming in is attached to an iterset and has already been dupped.

connorjward: Ah fantastic. This line was the reason for the small performance regression we were seeing so it's nice that it's not necessary. Will fix.

PR #630, opened 2021-07-13
Shortcut dat copy

dham:
Currently dat copying creates a par loop. This is slow and can dominate runtime on a small adjoint problem. In fact, numpy knows how to copy numpy arrays very quickly, so in the simplest case we should just do this.

This could possibly also be extended to the case where there is a subset.

### DIFF
--- a/pyop2/base.py
+++ b/pyop2/base.py
@@ -1629,6 +1629,8 @@ class Dat(DataCarrier, _EmptyDataMixin):
         :arg subset: A :class:`Subset` of entries to zero (optional)."""
         if subset is None:
             self.data[:] = 0
+        elif subset.superset != self.dataset.set:
+            raise MapValueError("The subset and dataset are incompatible")
         else:
             self.data[subset.owned_indices] = 0
 
@@ -1638,11 +1640,12 @@ class Dat(DataCarrier, _EmptyDataMixin):
 
         :arg other: The destination :class:`Dat`
         :arg subset: A :class:`Subset` of elements to copy (optional)"""
-        # breakpoint()
         if other is self:
             return
         if subset is None:
             other.data[:] = self.data_ro
+        elif subset.superset != self.dataset.set:
+            raise MapValueError("The subset and dataset are incompatible")
         else:
             other.data[subset.owned_indices] = self.data_ro[subset.owned_indices]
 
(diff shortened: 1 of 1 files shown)


pyop2/base.py
```diff
@@ -1579,6 +1579,9 @@ def copy(self, other, subset=None):
         :arg subset: A :class:`Subset` of elements to copy (optional)"""
         if other is self:
             return
+        if subset is None:
+            other.data[:] = self.data
```
wence-, line 1583: `self.data_ro`.

If you want to handle the subset case:
```
if subset is None:
    other.data[:] = self.data_ro
else:
    other.data[subset.indices] = self.data_ro[subset.indices]
```

wence-: I think you need to assert that the datasets for `other` and `self` match (and that if there's a subset that it also matches).

connorjward: If we do this then can we delete `_copy_parloop`?

connorjward:
Should we do the same for `zero` as well?

Melina97:
I ran into a type error so changed 'a += b' to np.add(a, b, out=a, casting='unsafe'), is that ok?

pyop2/base.py
```diff
@@ -1698,19 +1653,6 @@ def _iop_kernel(self, op, globalp, other_is_self, dtype):
...
-        globalp = False
-        if np.isscalar(other):
-            other = _make_object('Global', 1, data=other)
-            globalp = True
-        elif other is not self:
-            self._check_shape(other)
-        args = [self(INC)]
-        if other is not self:
-            args.append(other(READ))
-        par_loop(self._iop_kernel(op, globalp, other is self, other.dtype), self.dataset.set, *args)
```
connorjward, line 1711: Do we use `_iop_kernel` anywhere else? If not that can go.

pyop2/base.py
```diff
@@ -1825,19 +1767,41 @@ def __truediv__(self, other):
 
     def __iadd__(self, other):
         """Pointwise addition of fields."""
-        return self._iop(other, operator.iadd)
+        from numbers import Number
+        if other is not None:
+            if isinstance(other, Number):
+                self.data[:] += other
+            else:
+                np.add(self.data[:], other.data_ro, out=self.data[:], casting="unsafe")
```
connorjward, line 1775: I don't think we should be doing this as its behaviour is quite undefined. We should only add things if their shapes match. If you look in `_iop` above you'll see that we call `_check_shape`. Perhaps we should call that here.

Melina97: Yes, that sounds much better. Thanks!

dham:
OK I have Firedrake tests running on this now. @Melina97 can you check whether `_iop` can be deleted as @connorjward suggested. If that works, we can merge this.

Melina97:
I was prompted to do that last merge in order to push the changes, is that alright?

dham:
> I was prompted to do that last merge in order to push the changes, is that alright?

That part is fine, but now that Connor has fixed the builder, we can see that there are actual test fails that need looking at.

connorjward:
The tests that are failing suggest that being able to add things with different data types is actually what we want. I think setting `casting="unsafe"` but with `_check_shape` is probably the way to go.

pyop2/base.py
```diff
@@ -1907,19 +1817,45 @@ def __truediv__(self, other):
...
-        return self._iop(other, operator.itruediv)
+        from numbers import Number
+        if isinstance(other, Number):
+            self.data[:] /= other
+        else:
+            self._check_shape(other)
+            np.true_divide(self.data[:], other.data_ro, out=self.data[:], casting="unsafe")
+        return self
 
     __idiv__ = __itruediv__  # Python 2 compatibility
```
connorjward, line 1860: Can you delete this line at the same time? Python 2 is long gone.

connorjward (changes requested):
Just 2 comments. Looks good otherwise.

pyop2/base.py
```diff
@@ -1620,38 +1620,11 @@ def zero(self, subset=None):
...
-
-                knl = _make_object('Kernel', knl, 'zero')
-                self._zero_kernels[(self.dtype, self.cdim)] = knl
-            loop = _make_object('ParLoop', knl,
-                                iterset,
-                                self(WRITE))
-            loops[iterset] = loop
-        loop.compute()
+            self.data[subset.indices] = 0
+        return self
```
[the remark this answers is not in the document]
Melina97: Oh you're right, I'll remove that.

connorjward (approved):

connorjward:
I think that this is ready to go. The [upstream checks](https://github.com/firedrakeproject/firedrake/pull/2158) are now passing.

The changes I made from Melina's work are:
- Reverted the changes to `__iadd__`, `__imul__` etc. They were breaking for mixed dats and aren't a current bottleneck like `zero` and `copy` are. We could address it in a separate PR at some point.
- Created and used `subset.owned_indices` instead of `subset.indices`. The reason being we were getting strange failures due to trying to index elements inside the dat's halo.

pyop2/base.py
```diff
@@ -1620,38 +1627,12 @@ def zero(self, subset=None):
         """Zero the data associated with this :class:`Dat`
 
         :arg subset: A :class:`Subset` of entries to zero (optional)."""
-        if hasattr(self, "_zero_parloops"):
-            loops = self._zero_parloops
+        if subset is None:
+            self.data[:] = 0
```
wence-, line 1631: There's an optimisation available here where you set `self._data[:] = 0` which fills in the halo region, and then set the `halo_valid` property (or whatever it is called) to `True`.

pyop2/base.py
```diff
@@ -1661,28 +1642,12 @@ def copy(self, other, subset=None):
...
-            data = [loopy.GlobalArg("self", dtype=self.dtype, shape=(self.cdim,)),
-                    loopy.GlobalArg("other", dtype=other.dtype, shape=(other.cdim,))]
-            knl = loopy.make_function([domain], [insn], data, name="copy", target=loopy.CTarget(), lang_version=(2018, 2))
-
-            self._copy_kernel = _make_object('Kernel', knl, 'copy')
-        return _make_object('ParLoop', self._copy_kernel,
-                            subset or self.dataset.set,
-                            self(READ), other(WRITE))
+        if subset is None:
+            other.data[:] = self.data_ro
```
wence-, line 1646: Similar kind of comment applies here. if `self.halo_valid` then copy the full array including halo regions (and set `other.halo_valid`).

pyop2/base.py
connorjward: Done.

pyop2/base.py
connorjward: Done.

PR #632, opened 2021-07-21
Fix pyop2-clean

connorjward:
The existing solution did not work because we now cache our code in subdirectories (e.g. `2a/c2da0fe98e483fea7fa17cb352c822_p194525.c `).

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -501,29 +501,27 @@ def clear_cache(prompt=False):
     :arg prompt: if ``True`` prompt before removing any files
     """
     cachedir = configuration['cache_dir']
+
     if not os.path.exists(cachedir):
+        print("Cache directory could not be found")
         return
-
-    dirs = [os.path.join(cachedir, dir_) for dir_ in os.listdir(cachedir)]
-    ndirs = len(dirs)
-
-    if ndirs == 0:
+    if len(os.listdir(cachedir)) == 0:
         print("No cached libraries to remove")
         return
 
     remove = True
     if prompt:
-        user = input("Remove %d cached libraries from %s? [Y/n]: " % (ndirs, cachedir))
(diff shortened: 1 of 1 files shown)


pyop2/compilation.py
```diff
@@ -503,28 +504,26 @@ def clear_cache(prompt=False):
...
+            user = input("Remove %d cached libraries from %s? [Y/n]: " % (ndirs, cachedir))
 
         if user.lower() == 'n':
             remove = False
 
     if remove:
-        print("Removing %d cached libraries from %s" % (nfiles, cachedir))
-        [os.remove(f) for f in files]
+        print("Removing %d cached libraries from %s" % (ndirs, cachedir))
+        [shutil.rmtree(dir_) for dir_ in dirs]
```
wence-, line 527: Would it just be simpler to remove the `cachedir` directly?

connorjward: Yeah happy to do that instead. I chose to delete subdirectories because that's basically what this code did before. What's your preference?

wence-: I think just deleting `cachedir` since that is more future proof

connorjward: Done

sv2518:
I have only just seen this PR, sorry. I just wanted to suggest to change something else about the cache cleaning. I had an issue the other day, where I didn't have the access rights to remove the directories. It took me quite a while to realise that because the clean didn't tell me. I don't really know why the rights on those folder had changed and I guess it's not really a case that would happen often, but maybe we could check that the directories really have been removed, and throw an error if not?

connorjward:
Were you running `firedrake-clean` or `pyop2-clean`?

I've just checked with the new `pyop2-clean` from this PR and if we have a permissions issue it raises an exception the user sees.

If you mean `firedrake-clean` then we should definitely make sure that it informs the user if it didn't work.

PR #638, opened 2021-09-27
Bug fix empty extruded set with variable layers

stephankramer:
This would fail if the provided layers array has zero length (e.g.
for boundary facets on an interior parallel domain) as numpy does not
allow you to call min() on such an array.


connorjward:
Is it reasonable to have an `ExtrudedSet` with no layers? I think @wence- is better placed to review this.

stephankramer:
Yeah, if you run in parallel a ds integral creates an ExtrudedSet of boundary facets, but if you run with sufficient cores there will be some processes that do not have any boundary facets. Fair enough - I just followed github's suggestion

PR #643, opened 2021-11-04
Wence/fix/permuted map

wence-:
This fixes situations where the same base map was being used by two (or more) different `PermutedMap`s.

The name handling is bad, I will come round and fix that in a separate PR.

### DIFF
--- a/pyop2/codegen/builder.py
+++ b/pyop2/codegen/builder.py
@@ -30,32 +30,18 @@ class Map(object):
 
     __slots__ = ("values", "offset", "interior_horizontal",
                  "variable", "unroll", "layer_bounds",
-                 "prefetch", "permutation")
+                 "prefetch", "_pmap_count")
 
     def __init__(self, map_, interior_horizontal, layer_bounds,
-                 values=None, offset=None, unroll=False):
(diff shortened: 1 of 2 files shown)


pyop2/codegen/builder.py
```diff
@@ -112,26 +96,58 @@ def indexed(self, multiindex, layer=None):
...
+        self.offset = map_.offset
+        offset = map_.offset
+        # TODO: this is a hack, rep2loopy should be in charge of
+        # generating all names!
+        count = next(map_._pmap_count)
+        if offset is not None:
+            if offset.shape:
+                # Have a named literal
+                offset = offset.value[permutation]
+                offset = NamedLiteral(offset, parent=self.values, suffix=f"permutation{count}_offset")
```
sv2518, line 136: Maybe you can use the `UniqueNameGenerator` from pytools, probably inside the `NamedLiteral` class

wence-: I'm going to do a bigger refactor of the way nodes are named, I think. Since this is generically fragile.

wence-:
Firedrake passes: firedrakeproject/firedrake#2270

PR #644, opened 2022-01-26
Add disk_cached and cached decorators

connorjward:
As part of my quest to cache anything and everything in Firedrake I have added two new caching decorators to PyOP2:
- `@cached` which does in-memory caching only (I've stolen this directly from `cachetools`).
- `@disk_cached` which does in-memory and disk caching. I've largely stolen this from [Firedrake](https://github.com/firedrakeproject/firedrake/blob/master/firedrake/tsfc_interface.py#L73) but wrapped it up in a decorator.

### DIFF
--- a/pyop2/caching.py
+++ b/pyop2/caching.py
@@ -33,7 +33,15 @@
 
 """Provides common base classes for cached objects."""
 
+import hashlib
+import os
+from pathlib import Path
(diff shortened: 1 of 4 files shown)


dham (approved):
This looks good. Dealing with the parallel bit is future work.

PR #646, opened 2022-02-04
PyOP2 Logging: add PETSc events into the linear algebra callables.

sv2518:
I added logging into the C kernels for the solve and inverse (which are what Slate solves and inverses are translated to).

### DIFF
--- a/pyop2/codegen/c/inverse.c
+++ b/pyop2/codegen/c/inverse.c
@@ -1,6 +1,5 @@
 #include <petscsys.h>
 #include <petscblaslapack.h>
-#include <petsclog.h>
 
 #ifndef PYOP2_WORK_ARRAYS
 #define PYOP2_WORK_ARRAYS
@@ -9,33 +8,34 @@ static PetscBLASInt ipiv_buffer[BUF_SIZE];
 static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
 #endif
 
+#ifndef PYOP2_INV_LOG_EVENTS
+#define PYOP2_INV_LOG_EVENTS
+static PetscLogEvent USER_EVENT_inv_memcpy;
+static PetscLogEvent USER_EVENT_inv_getrf;
+static PetscLogEvent USER_EVENT_inv_getri;
+#endif
+
 static void inverse(PetscScalar* __restrict__ Aout, const PetscScalar* __restrict__ A, PetscBLASInt N)
 {
-    PetscLogEvent  USER_EVENT_memcpy;
-    PetscLogEvent  USER_EVENT_getrf;
-    PetscLogEvent  USER_EVENT_getri;
-    PetscClassId   classid;
+    PetscLogEventRegister("PyOP2InverseCallable_memcpy",PETSC_OBJECT_CLASSID,&USER_EVENT_inv_memcpy);
+    PetscLogEventRegister("PyOP2InverseCallable_getrf",PETSC_OBJECT_CLASSID,&USER_EVENT_inv_getrf);
+    PetscLogEventRegister("PyOP2InverseCallable_getri",PETSC_OBJECT_CLASSID,&USER_EVENT_inv_getri);
 
-    PetscClassIdRegister("PyOP2InverseCallable",&classid);
-    PetscLogEventRegister("PyOP2InverseCallable_memcpy",classid,&USER_EVENT_memcpy);
-    PetscLogEventRegister("PyOP2InverseCallable_getrf",classid,&USER_EVENT_getrf);
-    PetscLogEventRegister("PyOP2InverseCallable_getri",classid,&USER_EVENT_getri);
-
-    PetscLogEventBegin(USER_EVENT_memcpy,0,0,0,0);
+    PetscLogEventBegin(USER_EVENT_inv_memcpy,0,0,0,0);
     PetscBLASInt info;
     PetscBLASInt *ipiv = N <= BUF_SIZE ? ipiv_buffer : malloc(N*sizeof(*ipiv));
     PetscScalar *Awork = N <= BUF_SIZE ? work_buffer : malloc(N*N*sizeof(*Awork));
     memcpy(Aout, A, N*N*sizeof(PetscScalar));
-    PetscLogEventEnd(USER_EVENT_memcpy,0,0,0,0);
+    PetscLogEventEnd(USER_EVENT_inv_memcpy,0,0,0,0);
 
-    PetscLogEventBegin(USER_EVENT_getrf,0,0,0,0);
+    PetscLogEventBegin(USER_EVENT_inv_getrf,0,0,0,0);
     LAPACKgetrf_(&N, &N, Aout, &N, ipiv, &info);
-    PetscLogEventEnd(USER_EVENT_getrf,0,0,0,0);
+    PetscLogEventEnd(USER_EVENT_inv_getrf,0,0,0,0);
 
     if(info == 0){
-        PetscLogEventBegin(USER_EVENT_getri,0,0,0,0);
+        PetscLogEventBegin(USER_EVENT_inv_getri,0,0,0,0);
         LAPACKgetri_(&N, Aout, &N, ipiv, Awork, &N, &info);
-        PetscLogEventEnd(USER_EVENT_getri,0,0,0,0);
+        PetscLogEventEnd(USER_EVENT_inv_getri,0,0,0,0);
     }
 
     if(info != 0){
--- a/pyop2/codegen/c/inverse.c
+++ b/pyop2/codegen/c/inverse.c
@@ -1,5 +1,6 @@
 #include <petscsys.h>
 #include <petscblaslapack.h>
+#include <petsclog.h>
 
 #ifndef PYOP2_WORK_ARRAYS
 #define PYOP2_WORK_ARRAYS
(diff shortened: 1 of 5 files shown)


pyop2/codegen/c/inverse.c
```diff
@@ -10,14 +11,33 @@ static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
...
 {
+    PetscLogEvent  USER_EVENT_memcpy;
+    PetscLogEvent  USER_EVENT_getrf;
+    PetscLogEvent  USER_EVENT_getrs;
+    PetscClassId   classid;
+
+    PetscClassIdRegister("PyOP2InverseCallable",&classid);
+    PetscLogEventRegister("PyOP2InverseCallable_memcpy",classid,&USER_EVENT_memcpy);
+    PetscLogEventRegister("PyOP2InverseCallable_getrf",classid,&USER_EVENT_getrf);
+    PetscLogEventRegister("PyOP2InverseCallable_gerti",classid,&USER_EVENT_getrs);
```
connorjward, line 22: It's probably fine as it is, but we could wrap these in preprocessor macros so they are conditionally included. For example:
```C
#ifdef PYOP2_PROFILING
PetscClassIdRegister(...);
...
#endif
```
I think that this is very unlikely to be a performance problem though.

connorjward (commented):
Looks good! Any idea why tests are failing?

sv2518:
The tests were probably failing because I made a copy-paste error in the inverse callable.

PYOP2_PROFILING is automatically set when I run with the flamegraph option, right?

connorjward:
> The tests were probably failing because I made a copy-paste error in the inverse callable.

OK.

> PYOP2_PROFILING is automatically set when I run with the flamegraph option, right?

No it doesn't exist at the moment. In order to implement it you would need to pass `-DPYOP2_PROFILING` as an argument to the compiler. It would be nice to have but I don't think it should be a part of this PR.

Also, I'd be interested in whether we could do the same for the local kernel in a standard parloop. That would require actual code generation though which is less trivial.

sv2518:
Yeah this was easy because the callables are written in C anyways. I want to figure out how to do this for the local kernels coming from Slate. If I managed it out for them, we should be able to do it for the local kernels from TSFC too. You can write `CInstruction`s in Loopy and I **think** the easiest way to get this done would be to use those to include the PETSc instructions at the beginning and end of a local kernel for the logging. The other option might be to go over `CallInstructions`, but I need to remember myself if this would work or not.

connorjward:
> Yeah this was easy because the callables are written in C anyways. I want to figure out how to do this for the local kernels coming from Slate. If I managed it out for them, we should be able to do it for the local kernels from TSFC too. You can write `CInstruction`s in Loopy and I **think** the easiest way to get this done would be to use those to include the PETSc instructions at the beginning and end of a local kernel for the logging. The other option might be to go over `CallInstructions`, but I need to remember myself if this would work or not.

If you want to do local kernels I think the right place to insert the `PetscLogEventBegin/End` statements would be [here](https://github.com/OP2/PyOP2/blob/master/pyop2/codegen/builder.py#L908). You would obviously want the `PetscLogEventRegister` to be called at the beginning somewhere.

sv2518:
Ah yes, maybe it's much easier to do this from the outside in PyOP2 rather than from inside TSFC and Slate.

pyop2/codegen/c/inverse.c
sv2518: Okay since you said we can do that in a separate PR I opened an issue for it in #648 and will resolve this comment.

wence-:
FWIW, the logging infrastructure in PETSc is pretty lightweight, but it does add some non-trivial cost so I would probably say that logging per element kernel is likely a mistake.

connorjward:
Surely it would be acceptable to enable the local kernel logging as a configuration option that is false by default.

pyop2/codegen/c/inverse.c
```diff
@@ -10,14 +11,33 @@ static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
...
 {
+    PetscLogEvent  USER_EVENT_memcpy;
+    PetscLogEvent  USER_EVENT_getrf;
+    PetscLogEvent  USER_EVENT_getri;
+    PetscClassId   classid;
+
+    PetscClassIdRegister("PyOP2InverseCallable",&classid);
+    PetscLogEventRegister("PyOP2InverseCallable_memcpy",classid,&USER_EVENT_memcpy);
+    PetscLogEventRegister("PyOP2InverseCallable_getrf",classid,&USER_EVENT_getrf);
+    PetscLogEventRegister("PyOP2InverseCallable_getri",classid,&USER_EVENT_getri);
```
wence-, line 22: What we should do is make these events in Python once rather than every time inverse is called and then just get the handle in C.

pyop2/codegen/c/solve.c
```diff
@@ -10,17 +11,36 @@ static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
...
 {
+    PetscLogEvent  USER_EVENT_memcpy;
+    PetscLogEvent  USER_EVENT_getrf;
+    PetscLogEvent  USER_EVENT_getrs;
+    PetscClassId   classid;
+
+    PetscClassIdRegister("PyOP2SolveCallable",&classid);
+    PetscLogEventRegister("PyOP2SolveCallable_memcpy",classid,&USER_EVENT_memcpy);
+    PetscLogEventRegister("PyOP2SolveCallable_getrf",classid,&USER_EVENT_getrf);
+    PetscLogEventRegister("PyOP2SolveCallable_gerts",classid,&USER_EVENT_getrs);
```
wence-, line 22: This allocates memory every time around, so this should be done as something like:
```c
static PetscLogEvent USER_EVENT_memcpy;
...;
static void solve(...) {
    PetscLogEventBegin(...)
}
```
Then in Python:
```python
memcpy_event = PETSc.Log.Event("some_name")
dll = somehow_load_the_dll
setattr(dll, "USER_EVENT_memcpy", memcpy_event.handle)
```
I think

sv2518: Done in https://github.com/OP2/PyOP2/pull/646/commits/f55851f4ed6dedde3bd8adeb0e63c64c6f0e7da3

sv2518: Okay so now I am not only profiling the inverse and the solve here, but also the local TSFC and Slate kernels. The PRs for that are here https://github.com/firedrakeproject/tsfc/pull/267 and here https://github.com/firedrakeproject/firedrake/pull/2347. If I want to do something morally equivalent for those events too I need a better way to do this than I have now. I guess the options are
a) we do this generic by somehow collecting all variables declares as `PetscLogEvents` and then deduce the name of the event from the naming of the event id variable
b) we keep track of event ids and names somewhere.

sv2518:
Here is an example flame graph that showcases the new profiling features (of this PR and the related PRs in Slate and TSFC). In the example I have run a higher order (RTCF4-DG3) matrix-explicit hybridisation + GTMG preconditioner on a mixed Poisson problem. Now I can actually see in the profile that the matrix-explicit local linear algebra operations become critical
gtmg_matexpl_params_warmed_up_flame
.

pyop2/codegen/c/solve.c
connorjward: If we were only dealing with loopy kernels could we not inspect `kernel.instructions` (or similar, my loopy know-how is less than perfect) to find instances where we call `PetscLogEventRegister`? I don't know how this would work with the inverse and solve callables that you have though.

sv2518: If we decide to generate the Events in python we don't need to register the events in loopy anymore. In theory, we could do what you said with `PetscLogEventBegin` but at the point where we need the event name (here https://github.com/OP2/PyOP2/blob/f55851f4ed6dedde3bd8adeb0e63c64c6f0e7da3/pyop2/compilation.py#L504) there is no loopy anymore, only C. I think what we'd need to do in order to do this in a clean way is to save the event names and event id variable names on an attribute on the `LocalKernels` and then we could iterate over that to generate the events in `compilation.py`  because the `jitmodule` is a `GlobalKernel` and carries the `LocalKernel` (or `LocalKernels`??) from what I understand. This would be a bit of work because I need to return and save the information in all the right places, so I want to be sure first that we really want to do this. Let's postpone this discussion to the next Firedrake meeting.

pyop2/codegen/c/solve.c
```diff
@@ -8,19 +8,53 @@ static PetscBLASInt ipiv_buffer[BUF_SIZE];
 static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
 #endif
 
-static void solve(PetscScalar* __restrict__ out, const PetscScalar* __restrict__ A, const PetscScalar* __restrict__ B, PetscBLASInt N)
+#ifndef PYOP2_SOLVE_LOG_EVENTS
+#define PYOP2_SOLVE_LOG_EVENTS
+PetscLogEvent USER_EVENT_solve_memcpy = -1;
+PetscLogEvent USER_EVENT_solve_getrf = -1;
+PetscLogEvent USER_EVENT_solve_getrs = -1;
```
wence-, line 15: I think these want to be static? Or does that not work?

sv2518: Ahh I don't quite remember... I remember that I did have this static and think it was not working, but I wrestled with this stuff for a bit so maybe something else was the problem (rather than the static keyword)

sv2518:
Okay, I finally took the time to profile the difference in allocating the memory for the events in python and in C. For reproducibility I have a repo here https://github.com/sv2518/profiletheprofiling. The results are all gathered in this plot.

results

Here is what we see:

There are 4 setups: One setup with logging turned on, one with logging turned off. Both of these are run twice:
once for when we allocate the memory for the PETSc event in C (based on commit [7a68934](https://github.com/OP2/PyOP2/pull/646/commits/7a6893414d9311150cddcc7c65ffda44b1fc5416))
and once for when we generate the events in python (last commit on the branch).

We see the timing of a hybridisation solver run for lowest order RT-DG for 20 runs per setup (for more details see the repo). The average time spent per setup is indicated in the legend.

The critical question that I wanted to answer is if it is worth it to generate the events in python rather than in C. It's a bit of faff, in particular when we generate these events automatically in the production of slate kernels (other than the calls to inverse and solve).

Unfortunately it looks to me like we need to go through the faff, logging the inverse/solve callable with events which are allocated in C is noticably slower. The additional time of allocating the memory comes up in the flame graphs inside the Slate kernel time and is uncategorised so that's not great when I want to understand where all the time is spent in the local kernels.

In high order matrix explicit Slate kernels the overhead might be irrelevant because the local assembly and operations on it will be dominating the cost in the kernels, but I assume that for low order and theoretically high order matrix-free Slate kernels the overhead would be not negligible.

sv2518:
Closing in favour of https://github.com/OP2/PyOP2/pull/658

PR #647, opened 2022-02-04
Fix dtypes for COFFEE local kernel

connorjward:
It turns out that my method for determining the dtypes for a COFFEE local kernel was not robust so now it is treated the same as a `CStringLocalKernel`.

The reason for the magic that I do here is to enable backwards compatibility. `dtypes` is a property that 'belongs' to the local kernel but previously we would pass it in with the `Arg` objects. This means that the old `op2.Kernel` constructor does not accept this argument and we have to attach it later on.

This needs to get merged to fix Thetis tests ([related PR](https://github.com/thetisproject/thetis/pull/296)).

### DIFF
--- a/pyop2/local_kernel.py
+++ b/pyop2/local_kernel.py
@@ -211,25 +211,22 @@ class CStringLocalKernel(LocalKernel):
     def dtypes(self, dtypes):
         self._dtypes = dtypes
 
-    @cached_property
-    def arguments(self):
-        assert self.dtypes is not None
-
-        return tuple(LocalKernelArg(acc, dtype)
-                     for acc, dtype in zip(self.accesses, self.dtypes))
-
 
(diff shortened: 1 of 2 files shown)


pyop2/local_kernel.py
```diff
@@ -105,11 +105,6 @@ def __init__(self, code, name, accesses=None, *,
         self.requires_zeroed_output_arguments = requires_zeroed_output_arguments
         self.user_code = user_code
 
-    @property
-    @abc.abstractmethod
-    def dtypes(self):
```
JDBetteridge, line 110: Do we actually want to remove dtypes from the abstract base class here?
I'm thinking specifically of a scenario where a new kernel subclasses the ABC

JDBetteridge (changes requested):
I think a fix that doesn't change the ABC would avoid the need for similar fixes in the future

PR #650, opened 2022-02-18
Fix deadlocking disk_cached decorator

connorjward:
The existing `disk_cached` decorator deadlocks if `func` throws an exception. Here I rejig things to make sure that it is called collectively.


JDBetteridge (approved):
Looks good, `disk_cached` could easily be reworked into a class, but I'm not sure if it has a use case. Can revisit later if more features are desired

Issue #651, opened 2022-02-23
Deprecation warning for distutils Version classes.


mjg41:
Deprecation warning from PyOP2 from distutils in:
https://github.com/OP2/PyOP2/blob/master/pyop2/compilation.py

DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.

JDBetteridge:
Hmmm, `distutils` is [going away soon](https://www.python.org/dev/peps/pep-0632), we should probably try and remove all references to it in PyOP2

I've started work in a branch [here]() which should fix that issue. There are more references to `distutils` elsewhere that need removing.

Let me know if this solves your issue

PR #652, opened 2022-02-23
Remove distutils from compilation.py

JDBetteridge:
Address issue #651 and additionally try and remove `distutils` from PyOP2


pyop2/compilation.py
```diff
@@ -50,6 +50,29 @@
...
+        return "unknown"
+
+    def __gt__(self, other):
+        return False
+
+    def __eq__(self, other):
+        return False
+
+    def __lt__(self, other):
+        return False
```
connorjward, line 73: This seems like a massive hack. Does it not make more sense to define a `is_unknown_version` variable and then have
```py
if is_unknown_version:
    return
```
at the top of the necessary functions?

pyop2/compilation.py
```diff
@@ -81,23 +104,18 @@ def sniff_compiler_version(cc):
...
             ver = subprocess.check_output([cc, "-dumpversion"],
                                           stderr=subprocess.DEVNULL).decode("utf-8")
-            try:
-                ver = version.StrictVersion(ver.strip())
-            except ValueError:
-                # A sole digit, e.g. 7, results in a ValueError, so
-                # append a "do-nothing, but make it work" string.
-                ver = version.StrictVersion(ver.strip() + ".0")
-            if compiler == "gcc" and ver >= version.StrictVersion("7.0"):
+            ver = Version(ver.strip())
```
connorjward, line 112: Have you checked that a `ValueError` is no longer thrown for sole digits?

JDBetteridge: Yes
```python
>>> from packaging.version import Version
>>> v = Version('3')
>>> print(v)
3
>>> v < Version('3.4.5')
True
```

pyop2/compilation.py
JDBetteridge: Perhaps, I was just trying to recreate the behavior of `version.LooseVersion("unknown")`

JDBetteridge:
CI isn't running on this PR, I will open a new one

PR #654, opened 2022-03-01
Vectorisation sprint

sv2518:
Implements automatic cross-element vectorisation (This is the work from TJ and Kaushik)

See https://github.com/firedrakeproject/firedrake/pull/2365 for `firedrake` CI runs.
The corresponding loopy PR is https://github.com/inducer/loopy/pull/557.

Big thanks to @kaushikcfd for working hard on an update of this so that we can get it merged.

**The mechanism in PyOP2**

1. Check if kernel is vectorisable (see list below)
2. Change target of the kernel to `CVectorExtensionsTarget`
3. Inline all inner kernels in the wrapper kernel
4. Align all temporaries
5. Decide which instructions cannot be vectorised with extensions (see list below)
6. Shift the bound of the loop index (iname) to vectorise over so that it starts from 0
7. Split the iname (according to the SIMD length of the architecture which is determined with py-cpuinfo)
8. Break the temporaries (add a new axis to the temporary) and index it with the provided iname
9. Tag axes to vectorise over
10. Tag iname to vectorise with with `lp.VectorizeTag(lp.OpenMPSIMDTag())` where `VectorizeTag` indicates that we try to use vector extensions first, but if an instruction can't be vectorised we use the fallback `OpenMPSIMDTag` which wraps the instruction in openmp simd pragmas

**Kernels which cannot be vectorised**

- Kernels which assemble matrices
- Kernels which use complex types
- Kernels with read write access arguments
- The kernels which generate the extrusion coordinates
- Kernels with conditionals

**Single instructions which cannot be vectorised**
- Instructions which are outside the loop which was split (because they don't depend on the loop index we vectorise over)
- Constant literal temporaries on the RHS (because we cannot just index into them)
- Instruction with calls to Slate inverses and solve (gcc could do this, Kaushik extended the solve and inverse callables in PyOP2 with strided versions for that, but clang can't)

### DIFF
--- a/pyop2/configuration.py
+++ b/pyop2/configuration.py
@@ -44,16 +44,18 @@ def default_simd_width():
     from cpuinfo import get_cpu_info
     avx_to_width = {'avx': 2, 'avx1': 2, 'avx128': 2, 'avx2': 4,
                     'avx256': 4, 'avx3': 8, 'avx512': 8}
-    longest_ext = [t for t in get_cpu_info()["flags"] if t.startswith('avx')][-1]
-    if longest_ext not in avx_to_width.keys():
-        if longest_ext[:6] not in avx_to_width.keys():
-            assert longest_ext[:4] in avx_to_width.keys(), \
-                "The vector extension of your architecture is unknown. Disable vectorisation!"
-            return avx_to_width[longest_ext[:4]]
-        else:
-            return avx_to_width[longest_ext[:6]]
+    longest_simd_extension = [t for t in get_cpu_info()["flags"] if t.startswith('avx')][-1]
+    if longest_simd_extension in avx_to_width.keys():
+        return avx_to_width[longest_simd_extension]
+    elif longest_simd_extension[:6] in avx_to_width.keys():
+        return avx_to_width[longest_simd_extension[:6]]
+    elif longest_simd_extension[:4] in avx_to_width.keys():
+        return avx_to_width[longest_simd_extension[:4]]
     else:
-        return avx_to_width[longest_ext]
+        raise ConfigurationError(f"The vector extension of your architecture is unknown.\
+                                   Must be one of {str(avx_to_width.keys())}.\
+                                   We advise to disable vectorisation."
+                                 )
 
 
 class Configuration(dict):
--- a/pyop2/configuration.py
+++ b/pyop2/configuration.py
@@ -96,7 +96,7 @@ class Configuration(dict):
     :param alignment: A :class:`int` which specifies a size to which all temporaries
         are aligned in memory.
 
-        - ``sun2020study``: Cross-element vectorization strategy of
+        - ``cross-element``: Cross-element vectorization strategy of
           `<https://doi.org/10.1177/1094342020945005>`__.
     """
     # name, env variable, type, default, write once
@@ -117,7 +117,7 @@ class Configuration(dict):
         "simd_width":
             ("PYOP2_SIMD_WIDTH", int, default_simd_width()),
         "vectorization_strategy":
-            ("PYOP2_VECT_STRATEGY", str, "sun2020study"),
+            ("PYOP2_VECT_STRATEGY", str, "cross-element"),
         "alignment":
             ("PYOP2_ALIGNMENT", int, 64),
         "debug":
--- a/pyop2/configuration.py
+++ b/pyop2/configuration.py
@@ -54,7 +54,8 @@ def default_simd_width():
     else:
         raise ConfigurationError(f"The vector extension of your architecture is unknown.\
                                    Must be one of {str(avx_to_width.keys())}.\
-                                   We advise to disable vectorisation."
+                                   We advise to disable vectorisation \
+                                   with export PYOP2_VECT_STRATEGY=""."
                                  )
 
 
@@ -192,5 +193,7 @@ class Configuration(dict):
 
 
 configuration = Configuration()
+if configuration["vectorization_strategy"]:
+    configuration["simd_width"] = default_simd_width()
 
 target = CWithGNULibcTarget()
--- a/pyop2/configuration.py
+++ b/pyop2/configuration.py
@@ -95,11 +95,10 @@ class Configuration(dict):
     :param vectorization_strategy: A :class:`str` describing the
         vectorization strategy that must to be applied to the kernels. Can
         be one of the following --
-    :param alignment: A :class:`int` which specifies a size to which all temporaries
-        are aligned in memory.
-
         - ``cross-element``: Cross-element vectorization strategy of
           `<https://doi.org/10.1177/1094342020945005>`__.
+    :param alignment: A :class:`int` which specifies a size to which all temporaries
+        are aligned in memory.
     """
     # name, env variable, type, default, write once
     cache_dir = os.path.join(gettempdir(), "pyop2-cache-uid%s" % os.getuid())
--- a/pyop2/configuration.py
+++ b/pyop2/configuration.py
@@ -117,7 +117,7 @@ class Configuration(dict):
         "ldflags":
             ("PYOP2_LDFLAGS", str, ""),
         "simd_width":
-            ("PYOP2_SIMD_WIDTH", int, default_simd_width()),
+            ("PYOP2_SIMD_WIDTH", int, 1),
         "vectorization_strategy":
             ("PYOP2_VECT_STRATEGY", str, "cross-element"),
         "alignment":
--- a/pyop2/configuration.py
+++ b/pyop2/configuration.py
@@ -93,6 +93,8 @@ class Configuration(dict):
     :param vectorization_strategy: A :class:`str` describing the
         vectorization strategy that must to be applied to the kernels. Can
         be one of the following --
+    :param alignment: A :class:`int` which specifies a size to which all temporaries
+        are aligned in memory.
 
         - ``sun2020study``: Cross-element vectorization strategy of
           `<https://doi.org/10.1177/1094342020945005>`__.
@@ -118,8 +120,6 @@ class Configuration(dict):
             ("PYOP2_VECT_STRATEGY", str, "sun2020study"),
         "alignment":
             ("PYOP2_ALIGNMENT", int, 64),
-        "time":
-            ("PYOP2_TIME", bool, False),
         "debug":
             ("PYOP2_DEBUG", bool, False),
         "compute_kernel_flops":
--- a/pyop2/configuration.py
+++ b/pyop2/configuration.py
@@ -90,6 +90,12 @@ class Configuration(dict):
         cdim > 1 be built as block sparsities, or dof sparsities.  The
         former saves memory but changes which preconditioners are
         available for the resulting matrices.  (Default yes)
+    :param vectorization_strategy: A :class:`str` describing the
+        vectorization strategy that must to be applied to the kernels. Can
+        be one of the following --
+
+        - ``sun2020study``: Cross-element vectorization strategy of
+          `<https://doi.org/10.1177/1094342020945005>`__.
     """
     # name, env variable, type, default, write once
     cache_dir = os.path.join(gettempdir(), "pyop2-cache-uid%s" % os.getuid())
@@ -109,7 +115,7 @@ class Configuration(dict):
         "simd_width":
             ("PYOP2_SIMD_WIDTH", int, default_simd_width()),
         "vectorization_strategy":
-            ("PYOP2_VECT_STRATEGY", str, "ve"),
+            ("PYOP2_VECT_STRATEGY", str, "sun2020study"),
         "alignment":
             ("PYOP2_ALIGNMENT", int, 64),
         "time":
--- a/pyop2/global_kernel.py
+++ b/pyop2/global_kernel.py
@@ -380,22 +380,54 @@ class GlobalKernel(Cached):
 
         # align temps
         alignment = configuration["alignment"]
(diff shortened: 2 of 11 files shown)


pyop2/global_kernel.py
```diff
@@ -331,14 +333,68 @@ def code_to_compile(self):
         from pyop2.codegen.rep2loopy import generate
 
         wrapper = generate(self.builder)
+        if self._iterset._extruded:
+            iname = "layer"
+        else:
+            iname = "n"
+
+        has_matrix = any(arg._is_mat for arg in self._args)
```
connorjward, line 341: I think you want `isinstance(arg, MatKernelArg)` or something similar. I think `_is_mat` is gone.

pyop2/global_kernel.py
```diff
@@ -331,14 +333,68 @@ def code_to_compile(self):
...
+        has_rw = any(arg.access == op2.RW for arg in self._args)
+        is_cplx = any(arg.dtype.name == 'complex128' for arg in self._args)
```
connorjward, line 343: And for these I think you want `self.local_kernel.arguments` instead of `self._args`.

pyop2/global_kernel.py
```diff
@@ -331,14 +333,68 @@ def code_to_compile(self):
         from pyop2.codegen.rep2loopy import generate
 
         wrapper = generate(self.builder)
+        if self._iterset._extruded:
```
connorjward, line 336: I think just `self._extruded`

connorjward (commented):
I think there are going to be a few naming errors from my refactoring. I've labelled where they are.

pyop2/global_kernel.py
sv2518: Do we want to change that in https://github.com/OP2/PyOP2/blob/944c6cf73b302a1c6b2a5a4bc919db6818e8f96c/pyop2/global_kernel.py#L437 too?

pyop2/global_kernel.py
sv2518: I think I want `self.arguments`

sv2518:
All right this is passing now. :)

pyop2/global_kernel.py
```diff
@@ -439,7 +439,7 @@ def vectorise(self, wrapper, iname, batch_size):
         # }}}
 
         # split iname
-        slabs = (1, 1)
+        slabs = (0, 1)
```
sv2518, line 442: The loop always starts from 0 now because we have shifted the original loop index to `shifted_iname` right?

kaushikcfd: Yep!

sv2518:
This is ready for review. I will bring it up in the Firedrake meeting, but feel free to go ahead and review already.

pyop2/configuration.py
```diff
@@ -40,6 +40,22 @@
 from pyop2.exceptions import ConfigurationError
 
 
+def default_simd_width():
+    from cpuinfo import get_cpu_info
+    avx_to_width = {'avx': 2, 'avx1': 2, 'avx128': 2, 'avx2': 4,
+                    'avx256': 4, 'avx3': 8, 'avx512': 8}
+    longest_ext = [t for t in get_cpu_info()["flags"] if t.startswith('avx')][-1]
+    if longest_ext not in avx_to_width.keys():
+        if longest_ext[:6] not in avx_to_width.keys():
```
connorjward, line 49: This is very hard to read. Could you refactor to make more explicit?

pyop2/configuration.py
```diff
@@ -40,6 +40,22 @@
...
+            assert longest_ext[:4] in avx_to_width.keys(), \
```
connorjward, line 50: Users should never see assertion errors. Can you replace with [`ConfigurationError`](https://github.com/OP2/PyOP2/blob/master/pyop2/exceptions.py#L136) or, probably better, just fall back to a non-vectorised solution.

pyop2/configuration.py
```diff
@@ -74,6 +90,12 @@ class Configuration(dict):
         cdim > 1 be built as block sparsities, or dof sparsities.  The
         former saves memory but changes which preconditioners are
         available for the resulting matrices.  (Default yes)
+    :param vectorization_strategy: A :class:`str` describing the
+        vectorization strategy that must to be applied to the kernels. Can
+        be one of the following --
+
+        - ``sun2020study``: Cross-element vectorization strategy of
```
connorjward, line 97: Would it not be better to call this `"cross-element"`? Also this sort of configuration can be good to put into an `enum`.

pyop2/global_kernel.py
```diff
@@ -325,8 +326,60 @@ def builder(self):
...
+                       else dtype.is_complex()
+                       for dtype in self.local_kernel.dtypes
+                       )  # local args complex?
+                   or any(arg.dtype.is_complex()
+                          for arg in tuple(wrapper.default_entrypoint.temporary_variables.values())))  # global temps complex?
+        extruded_coords = self.local_kernel.name.endswith("extrusion")  # FIXME is there a better way to know that this kernel generated the extrusion coords?
+        vectorisable = ((not (has_matrix or has_rw)) and (configuration["vectorization_strategy"])) and not is_cplx and not extruded_coords
+
+        if vectorisable:
+            if isinstance(self.local_kernel.code, lp.TranslationUnit):
```
connorjward, line 354: Surely this line could just be added to the `vectorisable` predicate?

pyop2/configuration.py
```diff
@@ -91,7 +113,13 @@ class Configuration(dict):
...
             ("PYOP2_LDFLAGS", str, ""),
         "simd_width":
-            ("PYOP2_SIMD_WIDTH", int, 4),
+            ("PYOP2_SIMD_WIDTH", int, default_simd_width()),
+        "vectorization_strategy":
+            ("PYOP2_VECT_STRATEGY", str, "sun2020study"),
+        "alignment":
+            ("PYOP2_ALIGNMENT", int, 64),
+        "time":
+            ("PYOP2_TIME", bool, False),
```
connorjward, line 122: These configuration options are undocumented.

connorjward (changes requested):
I've left a couple of comments. Also, do we have a way to test this?

pyop2/configuration.py
sv2518: Since there is only the choice between `"cross-element"` and `""` for the vectorisation strategy, an enum is a bit overkill?

pyop2/global_kernel.py
```diff
@@ -336,6 +389,103 @@ def code_to_compile(self):
...
+        temps_not_to_vectorize |= {name
+                                   for name, tv in kernel.temporary_variables.items()
+                                   if (tv.read_only
+                                       and tv.initializer is not None)}
+
+        # {{{ clang (unlike gcc) does not allow taking address of vector-type
+        # variable
+
+        # FIXME: Perform this only if we know we are not using gcc.
+        for insn in kernel.instructions:
```
sv2518, line 440: This seems to be still failing if compiled with clang. Adding the temps in the call to `temps_not_to_vectorize` might not be sufficient.

sv2518:
Thanks for the review, Connor! We are testing this upstream in Firedrake and downstream in Loo.py. For testing in PyOP2 itself: vectorisation is turned on by default and there might be some tests in the PyOP2 test suite which are vectorisable, but I am not sure. I am not sure how I could test this more explicit in PyOP2.

requirements-ext.txt
```diff
@@ -4,6 +4,7 @@ pytest>=2.3
 flake8>=2.1.0
 pycparser>=2.10
 mpi4py>=1.3.1
+py-cpuinfo
```
JDBetteridge, line 7: Slightly concerned about adding this package as maintenance seems to have ceased. This could be problematic moving to the M1 Mac (https://github.com/workhorsy/py-cpuinfo/issues/162), but I might be wrong. I will try and look into an alternative and report back

sv2518: Aii, sorry. I had added this dependency two years ago and didn't look into it again. Thanks for trying to find an alternative, Jack!

pyop2/global_kernel.py
kaushikcfd: Thanks, fixed a typo in [a671b6c](https://github.com/OP2/PyOP2/pull/654/commits/a671b6ccb50e2ed7b12df1800534f99979bdb9fb). Looked at the generated code, looks fine now.

requirements-ext.txt
kaushikcfd: Doesn't look like there are great alternatives. I wouldn't mind, querying `platform` and hard-coding its [parameters](https://discussions.apple.com/thread/252073619).

sv2518:
@connorjward I added a test now. I also had to add the vectorisation config to the cache keys in [4aa86e1](https://github.com/OP2/PyOP2/pull/654/commits/4aa86e1414ab78d562698f8d5081a6c98b9d954b). I did this for all `def cache_key` functions in PyOP2, was that the right thing to do?

test/unit/test_vectorisation.py
```diff
@@ -0,0 +1,100 @@
...
+        # turn vectorisation off
+        op2.init(**{"vectorization_strategy": ""})
+
+        # Test that unvectorised code produced the correct result
+        ret = md.inner(md1)
+        assert abs(ret - 32) < 1e-12
+        ret = md1.inner(md)
+        assert abs(ret - 32) < 1e-12
+
+        # Test that we did not vectorise
```
sv2518, line 88: Also is there a better way to get the file to the generated code? This probably breaks if we run it in parallel.

test/unit/test_vectorisation.py
```diff
@@ -0,0 +1,100 @@
...
+    @pytest.fixture
+    def md(self, s):
+        n = op2.Dat(s, [4], np.float64)
+        o = op2.Dat(s, [5], np.float64)
+        md = op2.MixedDat([n, o])
+        return md
+
+    def test_vectorisation(self, md1, md):
+        # Test that vectorised code produced the correct result
+        ret = md.inner(md1)
```
connorjward, line 66: I feel like a better way to check the generated code would be to create a `Parloop` and then inspect `parloop.global_kernel.code_to_compile` ([link](https://github.com/OP2/PyOP2/blob/master/pyop2/parloop.py#L147), [link](https://github.com/OP2/PyOP2/blob/master/pyop2/global_kernel.py#L325)). You can look in `test/unit/test_indirect_loop.py` for an example but then you should use `LegacyParloop` instead of `par_loop` to get the right object back.

test/unit/test_vectorisation.py
connorjward: I feel like a better way to check the generated code would be to create a `Parloop` and then inspect `parloop.global_kernel.code_to_compile` ([link](https://github.com/OP2/PyOP2/blob/master/pyop2/parloop.py#L147), [link](https://github.com/OP2/PyOP2/blob/master/pyop2/global_kernel.py#L325)). You can look in `test/unit/test_indirect_loop.py` for an example but then you should use `LegacyParloop` instead of `par_loop` to get the right object back.

sv2518: In `test/unit/test_indirect_loop.py` all Kernels are generated from a string not from loopy. I didn't want to write a loopy kernel myself so I took a test where we are automatically generating a loopy kernel which we can vectorise already anyways. But I guess you are right because I have no access to the kernel that is he doing the calculations for `inner`

connorjward: Oh yes you're quite right. I think it would still be worth writing a loopy kernel so we can do this check without having to look at the filesystem.

sv2518: Actually I just unrolled now whatever was happening in the `inner` def on the dat, so I got around writing the kernel myself :)

pyop2/configuration.py
```diff
@@ -91,7 +117,11 @@ class Configuration(dict):
         "ldflags":
             ("PYOP2_LDFLAGS", str, ""),
         "simd_width":
-            ("PYOP2_SIMD_WIDTH", int, 4),
+            ("PYOP2_SIMD_WIDTH", int, default_simd_width()),
```
connorjward, line 120: If `default_simd_width` raises a `ConfigurationError` then won't this break?

connorjward (commented):
Only one comment. Otherwise looks good AFAICT. You should definitely run the Firedrake test suite to make sure you haven't broken anything by accident.

pyop2/configuration.py
sv2518: What do you mean with break?

I tried the following and it threw a `ConfigurationError` as expected?
```
from pyop2.exceptions import ConfigurationError
def myfunc(x): raise ConfigurationError("Can't use x")
mydict_of_tuples = {"simd": (1, 2, myfunc(1))}
```

sv2518: Ah actually I think I understand what you mean, it's breaking when people follow the advice in the configuration error and disable vectorisation this will still throw the error.

sv2518: Okay fixed

requirements-ext.txt
JDBetteridge: I'm 100% in favour of just hardcoding `PYOP2_SIMD_WIDTH=4` but for my alternative implementation of what is already implemented here see https://github.com/OP2/PyOP2/pull/665

connorjward:
Closing as https://github.com/OP2/PyOP2/pull/677 is a newer version of the same things.

PR #655, opened 2022-03-02
JDBetteridge/more compilers

JDBetteridge:
I want to add more compilers to PyOP2. Some of the structural changes come from [connorjward/add-non-forking-compiler](https://github.com/OP2/PyOP2/tree/connorjward/add-non-forking-compiler)

From here I want to add:
- Better support for the Intel compiler
- Clang compiler
- Cray compiler (which will supersede the hack in [JDBetteridge/isambard_fix](https://github.com/OP2/PyOP2/tree/JDBetteridge/isambard_fix)

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -59,7 +59,6 @@ def _check_hashes(x, y, datatype):
 
 
 _check_op = MPI.Op.Create(_check_hashes, commute=True)
-global _compiler
 _compiler = None
 
 
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -161,34 +161,35 @@ class Compiler(ABC):
 
     _cc = "mpicc"
     _cxx = "mpicxx"
+    _ld = None
 
-    _cppflags = None
-    _ldflags = None
+    _cflags = []
+    _cxxflags = []
+    _ldflags = []
 
-    _cflags = None
-    _cxxflags = None
-
-    _optflags = None
-    _debugflags = None
+    _optflags = []
+    _debugflags = []
 
     """A compiler for shared libraries.
     :arg cc: C compiler executable (can be overriden by exporting the
-        environment variable ``CC``).
+        environment variable ``PYOP2_CC``).
     :arg ld: Linker executable (optional, if ``None``, we assume the compiler
         can build object files and link in a single invocation, can be
-        overridden by exporting the environment variable ``LDSHARED``).
-    :arg cppargs: A list of arguments to the C compiler (optional, prepended to
-        any flags specified as the cflags configuration option)
-    :arg ldargs: A list of arguments to the linker (optional, prepended to any
-        flags specified as the ldflags configuration option).
+        overridden by exporting the environment variable ``PYOP2_LD``).
+    :arg extra_compiler_flags: A list of arguments to the C compiler (CFLAGS)
+        or the C++ compiler (CXXFLAGS)
+        ``PYOP2_CFLAGS`` and ``PYOP2_CXXFLAGS`` can also be used to extend these options
+        (optional, prepended to any flags specified as the cflags configuration option)
+    :arg extra_linker_flags: A list of arguments to the linker (LDFLAGS)
+        (optional, prepended to any flags specified as the ldflags configuration option).
     :arg cpp: Should we try and use the C++ compiler instead of the C
         compiler?.
     :kwarg comm: Optional communicator to compile the code on
         (defaults to COMM_WORLD).
     """
-    def __init__(self, extra_cppflags=None, extra_ldflags=None, cpp=False, comm=None):
-        self._extra_cppflags = extra_cppflags or []
-        self._extra_ldflags = extra_ldflags or []
+    def __init__(self, extra_compiler_flags=None, extra_linker_flags=None, cpp=False, comm=None):
+        self._extra_compiler_flags = extra_compiler_flags or []
+        self._extra_linker_flags = extra_linker_flags or []
 
         self._cpp = cpp
         self._debug = configuration["debug"]
@@ -199,37 +200,41 @@ class Compiler(ABC):
 
     @property
     def cc(self):
-        return os.environ.get("PYOP2_CC", self._cxx if self._cpp else self._cc)
+        return configuration["cc"] or self._cc
+
+    @property
+    def cxx(self):
+        return configuration["cxx"] or self._cxx
 
     @property
     def ld(self):
-        return os.environ.get("PYOP2_LD", None)
+        return configuration["ld"] or self._ld
 
     @property
-    def cppflags(self):
-        try:
-            return os.environ["PYOP2_CPPFLAGS"]
-        except KeyError:
-            cppflags = self._cppflags + self._extra_cppflags
+    def cflags(self):
+        cflags = self._cflags + self._extra_compiler_flags
+        if self._debug:
+            cflags += self._debugflags
+        else:
+            cflags += self._optflags
+        cflags += configuration["cflags"]
+        return cflags
 
-            if not self._debug:
-                cppflags += self._debugflags
-            else:
-                cppflags += self._optflags
-
-            if self._cpp:
-                cppflags += self._cxxflags
-            else:
-                cppflags += self._cflags
-
-            return cppflags
+    @property
+    def cxxflags(self):
+        cxxflags = self._cxxflags + self._extra_compiler_flags
+        if self._debug:
+            cxxflags += self._debugflags
+        else:
+            cxxflags += self._optflags
+        cxxflags += configuration["cxxflags"]
+        return cxxflags
 
     @property
     def ldflags(self):
-        try:
-            return os.environ["PYOP2_LDFLAGS"]
-        except KeyError:
-            return self._ldflags + self._extra_ldflags
+        ldflags = self._ldflags + self._extra_linker_flags
+        ldflags += configuration["ldflags"]
+        return ldflags
 
     @property
     def bugfix_cflags(self):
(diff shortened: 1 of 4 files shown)


pyop2/compilation.py
```diff
@@ -154,12 +155,23 @@ def compilation_comm(comm):
...
+    default_cxx = "mpicxx"
+
+    default_cppflags = None
+    default_ldflags = None
+
+    default_cflags = None
+    default_cxxflags = None
+
+    default_optflags = None
+    default_debugflags = None
```
connorjward, line 172: Maybe these should be empty tuples.

connorjward, line 172: Maybe these should be empty tuples.

pyop2/compilation.py
```diff
@@ -174,18 +186,54 @@ class Compiler(object):
...
-        self._cc = os.environ.get(ccenv, cc)
-        self._ld = os.environ.get('LDSHARED', ld)
-        self._cppargs = cppargs + configuration['cflags'].split()
-        if configuration["use_safe_cflags"]:
-            self._cppargs += self.workaround_cflags
-        self._ldargs = ldargs + configuration['ldflags'].split()
+
+    @property
+    def cc(self):
+        return os.environ.get("PYOP2_CC", self._cxx if self._cpp else self._cc)
```
connorjward, line 202: Probably a bad idea to mix C and C++ here.

pyop2/compilation.py
```diff
@@ -174,18 +186,54 @@ class Compiler(object):
...
+                cppflags += self._cflags
+
+            return cppflags
+
+    @property
+    def ldflags(self):
+        try:
+            return os.environ["PYOP2_LDFLAGS"]
+        except KeyError:
+            return self._ldflags + self._extra_ldflags
```
connorjward, line 232: I think I got a bit confused about this last time. Which flags do we consider essential and which are optional? Perhaps `self._ldflags` should be constant (i.e. the flags we will always want to have like `-lm`) and then the environment variable could only overwrite `_extra_ldflags`.

pyop2/compilation.py
```diff
@@ -227,20 +275,18 @@ def workaround_cflags(self):
...
-
         :arg jitmodule: The JIT Module which can generate the code to compile.
         :arg extension: extension of the source file (c, cpp).
-
         Returns a :class:`ctypes.CDLL` object of the resulting shared
         library."""
 
         # Determine cache key
-        hsh = md5(str(jitmodule.cache_key).encode())
+        hsh = md5(str(jitmodule.cache_key[1:]).encode())
```
connorjward, line 286: This is no longer the right thing to do.

pyop2/compilation.py
```diff
@@ -355,94 +401,321 @@ def get_so(self, jitmodule, extension):
...
+    _cppflags = None
+    _ldflags = None
 
-    :kwarg comm: Optional communicator to compile the code on (only
-        rank 0 compiles code) (defaults to COMM_WORLD).
+    _cflags = None
+    _cxxflags = None
+
+    _optflags = None
+    _debugflags = None
```
connorjward, line 418: Shouldn't these be deleted in favour of `default_xxxflags` in the `Compiler` class?

pyop2/compilation.py
```diff
@@ -355,94 +401,321 @@ def get_so(self, jitmodule, extension):
...
+            import pydffi
+        except ImportError:
+            raise ImportError("DragonFFI must be installed first")
+        ffi = pydffi.FFI()
+        ffi.compile(jitmodule.codetocompile)
+
+
+class TinyCCompiler(NonForkingCompiler):
+    def compile(self, jitmodule, _):
+        ...
```
connorjward, line 718: This should go.

pyop2/compilation.py
```diff
@@ -484,16 +757,16 @@ def __init__(self, code, argtypes):
...
         elif compiler == 'gcc':
-            compiler = LinuxCompiler(cppargs, ldargs, cpp=cpp, comm=comm)
+            compiler = LinuxGnuCompiler(cppargs, ldargs, cpp=cpp, comm=comm)
         else:
             raise CompilationError("Unrecognized compiler name '%s'" % compiler)
     elif platform.find('darwin') == 0:
-        compiler = MacCompiler(cppargs, ldargs, cpp=cpp, comm=comm)
+        compiler = MacClangCompiler(cppargs, ldargs, cpp=cpp, comm=comm)
     else:
         raise CompilationError("Don't know what compiler to use for platform '%s'" %
```
connorjward, line 771: Maybe make this a factory method/function? E.g `def load_compiler():`.

pyop2/global_kernel.py
```diff
@@ -363,7 +362,7 @@ def compile(self, comm):
                                 cppargs=cppargs,
                                 ldargs=ldargs,
                                 restype=ctypes.c_int,
-                                compiler=compiler,
+                                compiler=None,
```
connorjward, line 365: This is horrible. We only call `load` once so the API should be changed to avoid this.

connorjward (changes requested):
Thanks Jack for your work on this. `compilation.py` was really overdue an upgrade.

pyop2/compilation.py
```diff
@@ -481,24 +537,33 @@ def __init__(self, code, argtypes):
...
-    if platform.find('linux') == 0:
-        if compiler == 'icc':
-            compiler = LinuxIntelCompiler(cppargs, ldargs, cpp=cpp, comm=comm)
-        elif compiler == 'gcc':
-            compiler = LinuxCompiler(cppargs, ldargs, cpp=cpp, comm=comm)
+    cpp = (extension == "cpp")
+
+    # Sniff compiler from executable
+    exe = (configuration["cxx"] if cpp else configuration["cc"])
+    exe = exe or ("g++" if cpp else "gcc")
```
connorjward, line 544:
```py
if cpp:
    exe = configuration["cxx"] or "g++"
else:
    exe = configuration["cc"] or "gcc"
```
?
Also the logic seems a little bit tangled. We have an `exe` variable that will default to `"gcc"` that then gets transformed to `"GNU"`, `"Cray"` etc and *then* gets converted to `LinuxGnuCompiler`, `LinuxCrayCompiler` etc.Can this not be streamlined to `configuration["cc"] -> Linux{Gnu,Cray}Compiler`?

pyop2/compilation.py
```diff
@@ -355,99 +383,127 @@ def get_so(self, jitmodule, extension):
...
+    _cflags = ["-fPIC", "-Wall", "-framework", "Accelerate", "-std=gnu11"]
+    _cxxflags = ["-fPIC", "-Wall", "-framework", "Accelerate"]
+    _ldflags = ["-dynamiclib"]
 
-    def __init__(self, cppargs=[], ldargs=[], cpp=False, comm=None):
+    @property
+    def _optflags(self):
         machine = platform.uname().machine
         opt_flags = ["-O3", "-ffast-math"]
         if machine == "arm64":
```
connorjward, line 398: I feel like we should have separate Arm and x86 compiler classes so we don't need this here.

connorjward, line 398: I feel like we should have separate Arm and x86 compiler classes so we don't need this here.

pyop2/compilation.py
```diff
@@ -154,95 +157,121 @@ def compilation_comm(comm):
...
-            return Compiler.compiler_versions[key]
-        except KeyError:
-            if self.comm.rank == 0:
-                ver = sniff_compiler_version(self._cc)
-            else:
-                ver = None
-            ver = self.comm.bcast(ver, root=0)
-            return Compiler.compiler_versions.setdefault(key, ver)
+    def cc(self):
+        return configuration["cc"] or self._cc
```
connorjward, line 208: Surely this is now redundant since we inspect `configuration["cc"]` earlier when deciding which subclass of `Compiler` to instantiate.

connorjward, line 208: Surely this is now redundant since we inspect `configuration["cc"]` earlier when deciding which subclass of `Compiler` to instantiate.

pyop2/compilation.py
```diff
@@ -355,99 +383,127 @@ def get_so(self, jitmodule, extension):
...
-        super(LinuxCompiler, self).__init__(cc, cppargs=cppargs, ldargs=ldargs,
-                                            cpp=cpp, comm=comm)
+    _debugflags = ["-O0", "-g"]
+
+
+class LinuxGnuCompiler(Compiler):
+    """The GNU compiler for building a shared library on Linux systems."""
+    _name = "GNU"
+
+    _cflags = ["-fPIC", "-Wall", "-std=gnu11"]
```
connorjward, line 412: I'm not a big fan of these being private attributes but it's probably the best solution.

connorjward, line 412: I'm not a big fan of these being private attributes but it's probably the best solution.

connorjward (changes requested):

pyop2/compilation.py
JDBetteridge: I think this is the right thing to do because `PYOP2_CC` would normally be set to something like `/opt/mpich/bin/mpicc` and there's not a direct route from this to the right compiler class without doing something like what `sniff_compiler_version` does.

The Cray compiler and ScoreP compilers spring to mind as examples of compilers that require some sniffing to work out what they really are.

This does raise the question of whether `gcc` is the correct default here though.

pyop2/compilation.py
JDBetteridge: I think you still want the ability to override this, for example, the Intel compiler is distributed as both `mpicc` and `mpiicc`. In the latter case `mpicc` is actually a wrapper for the GNU compiler!

pyop2/compilation.py
JDBetteridge: I'm happy to take on an alternative suggestion, this is mainly so the cflags getter method works correctly. One alternative is hard coding them into the methods, but I'm even less of a fan of that!

pyop2/compilation.py
```diff
@@ -285,65 +370,66 @@ def get_so(self, jitmodule, extension):
...
-                                                              stdout=log)
-                                except subprocess.CalledProcessError as e:
-                                    raise CompilationError(
-                                        """Command "%s" return error status %d.
+                        with open(logfile, "w") as log, open(errfile, "w") as err:
+                            log.write("Compilation command:\n")
+                            log.write(" ".join(cc))
+                            log.write("\n\n")
+                            try:
+                                if configuration['no_fork_available']:
```
connorjward, line 384: This configuration option is another one of the meaningless ones so should probably get deleted.

connorjward (approved):
Looks good.

pyop2/compilation.py
JDBetteridge: I was under the impression that on a system that prohibited forking, this would (at least in some cases) allow you to invoke the compiler.

PR #658, opened 2022-03-24
Profile local kernels and set events on dll

sv2518:
Supercedes #646
Related to [#2388](https://github.com/firedrakeproject/firedrake/pull/2388/files) and [#273](https://github.com/firedrakeproject/tsfc/pull/273/files)

TSFC and Slate generate instructions for logging the local kernels. Both TSFC and Slate generate event names and set event ids to a default value. Now in PyOP2 we generate the actual event and link the id of it to the id in the dllwhich contains the local kernel (as replacement for the default value).

### DIFF
--- a/pyop2/codegen/c/inverse.c
+++ b/pyop2/codegen/c/inverse.c
@@ -8,16 +8,56 @@ static PetscBLASInt ipiv_buffer[BUF_SIZE];
 static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
 #endif
 
+#ifndef PYOP2_INV_LOG_EVENTS
+#define PYOP2_INV_LOG_EVENTS
+static PetscLogEvent USER_EVENT_inv_memcpy;
+static PetscLogEvent USER_EVENT_inv_getrf;
+static PetscLogEvent USER_EVENT_inv_getri;
+#endif
+
+#ifndef BEGIN_LOG
+#define BEGIN_LOG
+static void beginLog(PetscLogEvent eventId){
+    #ifdef PYOP2_PROFILING_ENABLED
+    PetscLogEventBegin(eventId,0,0,0,0);
+    #endif
+}
+#endif
+
+#ifndef END_LOG
+#define END_LOG
+static void endLog(PetscLogEvent eventId){
+    #ifdef PYOP2_PROFILING_ENABLED
+    PetscLogEventEnd(eventId,0,0,0,0);
+    #endif
+}
+#endif
+
 static void inverse(PetscScalar* __restrict__ Aout, const PetscScalar* __restrict__ A, PetscBLASInt N)
 {
+    #ifdef PYOP2_PROFILING_ENABLED
+    PetscLogEventRegister("PyOP2InverseCallable_memcpy",PETSC_OBJECT_CLASSID,&USER_EVENT_inv_memcpy);
+    PetscLogEventRegister("PyOP2InverseCallable_getrf",PETSC_OBJECT_CLASSID,&USER_EVENT_inv_getrf);
+    PetscLogEventRegister("PyOP2InverseCallable_getri",PETSC_OBJECT_CLASSID,&USER_EVENT_inv_getri);
+    #endif
+
+    beginLog(USER_EVENT_inv_memcpy);
     PetscBLASInt info;
     PetscBLASInt *ipiv = N <= BUF_SIZE ? ipiv_buffer : malloc(N*sizeof(*ipiv));
     PetscScalar *Awork = N <= BUF_SIZE ? work_buffer : malloc(N*N*sizeof(*Awork));
     memcpy(Aout, A, N*N*sizeof(PetscScalar));
+    endLog(USER_EVENT_inv_memcpy);
+
+    beginLog(USER_EVENT_inv_getrf);
     LAPACKgetrf_(&N, &N, Aout, &N, ipiv, &info);
+    endLog(USER_EVENT_inv_getrf);
+
     if(info == 0){
+        beginLog(USER_EVENT_inv_getri);
         LAPACKgetri_(&N, Aout, &N, ipiv, Awork, &N, &info);
+        endLog(USER_EVENT_inv_getri);
     }
+
     if(info != 0){
         fprintf(stderr, "Getri throws nonzero info.");
         abort();
--- a/pyop2/codegen/c/inverse.c
+++ b/pyop2/codegen/c/inverse.c
@@ -10,52 +10,30 @@ static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
 
 #ifndef PYOP2_INV_LOG_EVENTS
 #define PYOP2_INV_LOG_EVENTS
-static PetscLogEvent USER_EVENT_inv_memcpy;
-static PetscLogEvent USER_EVENT_inv_getrf;
-static PetscLogEvent USER_EVENT_inv_getri;
+PetscLogEvent ID_inv_memcpy = -1;
+PetscLogEvent ID_inv_getrf = -1;
+PetscLogEvent ID_inv_getri = -1;
+static PetscBool log_active_inv = 0;
 #endif
 
-#ifndef BEGIN_LOG
-#define BEGIN_LOG
-static void beginLog(PetscLogEvent eventId){
-    #ifdef PYOP2_PROFILING_ENABLED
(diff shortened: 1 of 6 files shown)


sv2518:
@connorjward could you have a look at this? This is basically the follow-up code of our discussion on https://github.com/OP2/PyOP2/pull/646

sv2518:
I will still provide the measurement which proves that this is necessary fo profiling the local kernels without too much overhead. The plot in the other PR is only for when we shifting generating the events in C vs Python for the inverse and solve callable.

sv2518:
Okay so here the results, same explanation as for the plot on the other PR, but I profiled a different problem and solver (where more local kernels are generated). The improvement is essentially from blue to green.

results

pyop2/codegen/loopycompat.py
```diff
@@ -106,7 +106,9 @@ def _shape_1_if_empty(shape_caller, shape_callee):
 
             elif isinstance(callee_insn, (CInstruction,
                     _DataObliviousInstruction)):
-                pass
+                # Matching for c instructions is not ensure
+                # but just dropping them is no option
+                new_callee_insns.append(callee_insn)
```
connorjward, line 111: Sorry I don't understand this comment. Could you make it clearer?

pyop2/codegen/c/inverse.c
```diff
@@ -8,16 +8,34 @@ static PetscBLASInt ipiv_buffer[BUF_SIZE];
 static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
 #endif
 
-static void inverse(PetscScalar* __restrict__ Aout, const PetscScalar* __restrict__ A, PetscBLASInt N)
+#ifndef PYOP2_INV_LOG_EVENTS
+#define PYOP2_INV_LOG_EVENTS
+PetscLogEvent ID_inv_memcpy = -1;
+PetscLogEvent ID_inv_getrf = -1;
+PetscLogEvent ID_inv_getri = -1;
+static PetscBool log_active_inv = 0;
```
connorjward, line 16: Does this need to be defined outside of the function body? I guess it prevents a tiny amount of memory allocation.

pyop2/compilation.py
```diff
@@ -185,6 +186,8 @@ def __init__(self, cc, ld=None, cppargs=[], ldargs=[],
         self._cppargs = cppargs + configuration['cflags'].split()
         if configuration["use_safe_cflags"]:
             self._cppargs += self.workaround_cflags
+        if PETSc.Log.isActive():
+            self._cppargs += ["-DPYOP2_PROFILING_ENABLED"]
```
connorjward, line 190: Is this option still required? I don't see it in the C code anywhere.

pyop2/local_kernel.py
```diff
@@ -92,7 +92,8 @@ def __init__(self, code, name, accesses=None, *,
                  ldargs=(),
                  opts=None,
                  requires_zeroed_output_arguments=False,
-                 user_code=""):
+                 user_code="",
+                 events=()):
```
connorjward, line 96: Can this be added to the docstring?

pyop2/compilation.py
```diff
@@ -498,13 +501,34 @@ def __init__(self, code, argtypes):
...
+    if PETSc.Log.isActive():
+        # also link the events from the linear algebra callables
+        if hasattr(dll, "solve"):
+            events += ('solve_memcpy', 'solve_getrf', 'solve_getrs')
+        if hasattr(dll, "inverse"):
+            events += ('inv_memcpy', 'inv_getrf', 'inv_getri')
+        # link all ids in DLL to the events generated here in python
+        for e in events:
+            ctypes.c_int.in_dll(dll, 'ID_'+e).value = PETSc.Log.Event(e).id
+    return dll
```
connorjward, line 529: I would call this an anti-pattern (possibly I'm wrong). If you are modifying `dll` in-place it should not also be returned from the function.

pyop2/compilation.py
```diff
@@ -498,13 +501,34 @@ def __init__(self, code, argtypes):
         raise CompilationError("Don't know what compiler to use for platform '%s'" %
                                platform)
     dll = compiler.get_so(code, extension)
+    if code.local_kernel.events:
```
connorjward, line 504: Is this check valid? If `events` is still empty we still may want to profile `solve` and `inverse`.

connorjward (changes requested):
Thanks Sophia. I really like what you've done. I just have some minor comments.

pyop2/codegen/c/inverse.c
sv2518: Yes it doesn't need to be but it can so I thought I'd do that.

pyop2/compilation.py
sv2518: No, it's not will remove

pyop2/compilation.py
sv2518: Yes will change it

pyop2/compilation.py
sv2518: Solve and inverse need to be called on some matrix and the matrix needs to be filled with data through a local assembly kernel. So the events variable would never be empty if inverse and solve are called/want to be profiled. But I will change it anyways, so that someone else does not have the same thoughts as you later.

pyop2/codegen/c/inverse.c
```diff
@@ -8,16 +8,34 @@ static PetscBLASInt ipiv_buffer[BUF_SIZE];
...
+#endif
+
+void inverse(PetscScalar* __restrict__ Aout, const PetscScalar* __restrict__ A, PetscBLASInt N)
 {
+    PetscLogIsActive(&log_active_inv);
+    if (log_active_inv){PetscLogEventBegin(ID_inv_memcpy,0,0,0,0);}
```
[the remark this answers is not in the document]
sv2518: Maybe not but I prefer if they are there

pyop2/codegen/c/inverse.c
```diff
@@ -8,16 +8,34 @@ static PetscBLASInt ipiv_buffer[BUF_SIZE];
 static PetscScalar work_buffer[BUF_SIZE*BUF_SIZE];
 #endif
 
-static void inverse(PetscScalar* __restrict__ Aout, const PetscScalar* __restrict__ A, PetscBLASInt N)
+#ifndef PYOP2_INV_LOG_EVENTS
+#define PYOP2_INV_LOG_EVENTS
+PetscLogEvent ID_inv_memcpy = -1;
+PetscLogEvent ID_inv_getrf = -1;
+PetscLogEvent ID_inv_getri = -1;
```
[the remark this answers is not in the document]
sv2518: I did that for debugging purposes (I printed the event ids in the kernels to see that they are actually set. It was easier to check by setting a default value). If the initialisation has no disadvantage I'd prefer to keep it.

pyop2/local_kernel.py
sv2518: Jup!

pyop2/compilation.py
connorjward: It doesn't look like you've applied the change.

pyop2/compilation.py
```diff
@@ -498,13 +499,34 @@ def __init__(self, code, argtypes):
         raise CompilationError("Don't know what compiler to use for platform '%s'" %
                                platform)
     dll = compiler.get_so(code, extension)
+    if isinstance(jitmodule, GlobalKernel):
```
connorjward, line 502: Is there ever a case when this is `False`? I'm pretty sure that `jitmodule` is always a `GlobalKernel`.

sv2518: I am pretty sure it is not always a GlobalKernel, see https://github.com/OP2/PyOP2/blob/42812fad033509fdf1a77cf34ed82828f757d75c/pyop2/compilation.py#L471

pyop2/compilation.py
sv2518: Argh, my bad, I must have dropped my change when I squashed.

pyop2/compilation.py
connorjward: Ah OK.

pyop2/compilation.py
sv2518: Okay changed now.

connorjward (approved):

sv2518:
Thanks for the review, Connor!

PR #659, opened 2022-03-28
Add dtype check for parloops

connorjward:
This should resolve https://github.com/firedrakeproject/firedrake/discussions/2391.


wence- (approved):
I think that looks right.

PR #663, opened 2022-05-04
[loopycompat.Mappers]: call super's constructor

kaushikcfd:
There are some potential updates to Loopy's expression traversals in https://github.com/inducer/loopy/pull/603. The downstream CI in that PR caught that we weren't calling the super-class's constructor in the PyOP2's mapper implementations. This patch should fix it.

/cc @inducer


wence- (approved):
Thanks!

PR #664, opened 2022-05-06
PyOP2 compilation: add a pathway to compile with gcc on Mac.

sv2518:
I tried to compile with `PYOP2_CC=gcc-9` but the logic in `compilation.py` made me end up in the `AnonymousCompiler` which does not set any linker flags. I don't want to set these by hand so I added a `MacGNUCompiler` (which uses the the same flags as the clang compiler). (Thanks @JDBetteridge)


connorjward (approved):
Looks fine to me.

PR #665, opened 2022-05-27
Don't add py-cpuinfo

JDBetteridge:
This is just an idea for how to avoid using the `py-cpuinfo` package.

To completely honest I'm not sure this is even necessary:
- It will only work on x86_64 architectures (this is also true of `py-cpuinfo`)
- It feels hacky
- The default of `PYOP2_SIMD_WIDTH=4` covers almost all use cases (with the exception of AVX512*)

If a user _really_ wants to change the SIMD width used within PYOP2 they should manually set `PYOP2_SIMD_WIDTH`, I don't really think we should try and sniff this value.

*My impression is that using AVX512 normally incurs a performance penalty due to the CPU having to downclock. A benchmark might convince me otherwise.

### DIFF
--- a/pyop2/__init__.py
+++ b/pyop2/__init__.py
@@ -7,3 +7,8 @@ from pyop2.version import __version_info__  # noqa: just expose
 from pyop2._version import get_versions
 __version__ = get_versions()['version']
 del get_versions
+
+from pyop2.configuration import configuration
+from pyop2.compilation import max_simd_width
+if configuration["vectorization_strategy"]:
+    configuration["simd_width"] = max_simd_width()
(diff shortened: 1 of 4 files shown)


connorjward:
I agree. I think defaulting to 4, and allowing users to change things, is a completely reasonable approach here.

Are there ever circumstances where this might cause a breakage?

wence-:
> I agree. I think defaulting to 4, and allowing users to change things, is a completely reasonable approach here.
> 
> Are there ever circumstances where this might cause a breakage?

M1 cores have 128bit vectors. But I think that going for four wide (effectively 256bit vectors) would be fine.

sv2518:
Thanks, Jack!

PR #668, opened 2022-07-10
Refactor num_flops and fix for loopy kernels

sv2518:
This a fix for the flop counting. If we turn flop counting on in Firedrake without this fix, all tests are failing see actions run on this commit https://github.com/firedrakeproject/firedrake/pull/2490/commits/fae62c404eeb5175b0acccbfb259c8487ac34564. The commit later pulls this PyOP2 branch in.
I also refactored the code a tiny bit and made `num_flops` a method of the specialized local kernels rather than the generic to avoid some of the `isinstance` special casing we had before.
The coffee flop counting is fixed now too.

There is something funny going on with layer args in the flop counting algorithm that I don't understand yet (see failing tests @ https://github.com/firedrakeproject/firedrake/runs/7272623399?check_suite_focus=true. (EDIT: I had to fix the layer arg in the loopy kernels because we usually pass a value for that in from the global kernel)

Order of merging: https://github.com/firedrakeproject/tsfc/pull/279, then this one, then https://github.com/firedrakeproject/firedrake/pull/2490

### DIFF
--- a/pyop2/local_kernel.py
+++ b/pyop2/local_kernel.py
@@ -10,7 +10,6 @@ from loopy.tools import LoopyKeyBuilder
 import numpy as np
 
 from pyop2 import version
-from pyop2.configuration import configuration
 from pyop2.datatypes import ScalarType
 from pyop2.exceptions import NameTypeError
 from pyop2.types import Access
@@ -224,8 +223,6 @@ class CoffeeLocalKernel(LocalKernel):
     def num_flops(self):
         if self.flop_count is not None:
             return self.flop_count
-        elif not configuration["compute_kernel_flops"]:
-            return 0
         else:
             v = EstimateFlops()
             return v.visit(self.code)
@@ -254,8 +251,6 @@ class LoopyLocalKernel(LocalKernel):
     def num_flops(self):
         if self.flop_count is not None:
             return self.flop_count
-        elif not configuration["compute_kernel_flops"]:
-            return 0
         else:
             if isinstance(self.code, lp.TranslationUnit):
                 prog = self.code.with_entrypoints(self.name)
--- a/pyop2/local_kernel.py
+++ b/pyop2/local_kernel.py
@@ -197,6 +197,8 @@ class CStringLocalKernel(LocalKernel):
 
     @cached_property
     def num_flops(self):
+        """Set the numbers of FLOPs to 0 if not already known,
+           because there is no way to measure or estimate the FLOPS for string kernels. """
         if self.flop_count is not None:
             return self.flop_count
         else:
@@ -221,6 +223,8 @@ class CoffeeLocalKernel(LocalKernel):
 
     @cached_property
     def num_flops(self):
+        """Compute the numbers of FLOPs if not already known
+           using COFFEE's FLOP estimation algorithm."""
         if self.flop_count is not None:
             return self.flop_count
         else:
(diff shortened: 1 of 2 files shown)


pyop2/local_kernel.py
```diff
@@ -231,6 +220,16 @@ def dtypes(self):
     def dtypes(self, dtypes):
         self._dtypes = dtypes
 
+    @cached_property
+    def num_flops(self):
+        if self.flop_count is not None:
+            return self.flop_count
+        elif not configuration["compute_kernel_flops"]:
```
sv2518, line 227: I was wondering if we can pull this if clause into the `_compute` function where we call the logging of the flops to avoid having this in every specialised kernel, what do you think @connorjward ?

sv2518:
Currently the counting is spewing a lot of warnings in loopy and silencing doesn't work as I would expect (possibly my fault). I asked the Loo.py people here https://github.com/inducer/loopy/issues/645.

pyop2/local_kernel.py
```diff
@@ -250,3 +246,25 @@ def _loopy_arguments(self):
...
+                warnings = list(knl.silenced_warnings)
+                warnings.extend(['insn_count_subgroups_upper_bound',
+                                 'get_x_map_guessing_subgroup_size',
+                                 'summing_if_branches_ops'])
+                knl = knl.copy(silenced_warnings=warnings,
+                               options=lp.Options(ignore_boostable_into=True))
+                knl = lp.fix_parameters(knl, layer=1)
+                prog = prog.with_kernel(knl)
+            else:
+                prog = self.code
```
connorjward, line 267: Can you explain what is going on here?

pyop2/local_kernel.py
```diff
@@ -152,28 +152,9 @@ def arguments(self):
                      for acc, dtype in zip(self.accesses, self.dtypes))
 
     @cached_property
+    @abc.abstractmethod
     def num_flops(self):
```
connorjward, line 156: Can the docstring explanation for `flop_count` be improved to say something like "...if this is not provided then PyOP2 will try to estimate a FLOP count.".

connorjward (changes requested):

pyop2/local_kernel.py
sv2518: Does PyOP2 do that?

sv2518: For string kernels it returns 0, for coffee kernels it returns an estimation and for loopy kernels it actually measures the FLOPS. Is there another type of local kernel?

pyop2/local_kernel.py
sv2518: I added my commit comments as actual comments now to make it more clear. I also think that actually self.code might always be a translation unit, I just wasn't sure because in the constructor we say that
`@validate_type(("code", (lp.LoopKernel, lp.TranslationUnit), TypeError))`. I changed this and will run all firedrake tests on this to make sure there is no special case.

pyop2/local_kernel.py
connorjward: No you're right. What I'm saying is that the docstring should describe what happens if `flop_count` is `None` as its behaviour is not obvious.

sv2518: Okay I added what I said here to the method description.

connorjward (approved):

sv2518:
Thanks, Connor! I will waiting with merging this until the warnings are shut up properly.

connorjward:
Closing as we are moving PyOP2 into Firedrake. There is a [Firedrake PR](https://github.com/firedrakeproject/firedrake/pull/2490) that references this.

PR #670, opened 2022-07-25
Fix Vec's state counter for MixedDat

nbouziani:
`MixedDat` objects don’t only rely on the PETSc Vec of the underlying `Dat` objects, they is also a Vec living on the mixed space that is made by duplicating the layout vector of the dataset attached to the `MixedDat` instance.

The data version of `MixedDat` is obtained by summing the data version of the underlying `Dat` (not unique but enough to detect that the data value changed), but we also need to update the data version of the vector in the mixed space in the `vec_context` where it is populated.

This causes a test to fail in firedrake as PETSc Vecs cache some operations (e.g. norm computations) based on the Vec's state counter which can lead to wrong numerical values.

### DIFF
--- a/pyop2/types/dat.py
+++ b/pyop2/types/dat.py
@@ -1015,13 +1015,12 @@ class MixedDat(AbstractDat, VecAccessMixin):
         # values
         if access is not Access.WRITE:
             offset = 0
-            array = self._vec.array
-            for d in self:
-                with d.vec_ro as v:
-                    size = v.local_size
-                    array[offset:offset+size] = v.array_r[:]
-                    offset += size
-            del array
+            with self._vec as array:
+                for d in self:
+                    with d.vec_ro as v:
+                        size = v.local_size
+                        array[offset:offset+size] = v.array_r[:]
+                        offset += size
 
         yield self._vec
         if access is not Access.READ:
--- a/pyop2/types/dat.py
+++ b/pyop2/types/dat.py
@@ -1021,6 +1021,8 @@ class MixedDat(AbstractDat, VecAccessMixin):
                     size = v.local_size
                     array[offset:offset+size] = v.array_r[:]
                     offset += size
+
+        self._vec.stateIncrease()
         yield self._vec
         if access is not Access.READ:
             # Reverse scatter to get the values back to their original locations
--- a/pyop2/types/dat.py
+++ b/pyop2/types/dat.py
@@ -1021,8 +1021,8 @@ class MixedDat(AbstractDat, VecAccessMixin):
                     size = v.local_size
                     array[offset:offset+size] = v.array_r[:]
                     offset += size
+            del array
 
-        self._vec.stateIncrease()
         yield self._vec
         if access is not Access.READ:
             # Reverse scatter to get the values back to their original locations


pyop2/types/dat.py
```diff
@@ -1021,6 +1021,8 @@ def vec_context(self, access):
                     size = v.local_size
                     array[offset:offset+size] = v.array_r[:]
                     offset += size
+
+        self._vec.stateIncrease()
```
wence-, line 1025: This should not be necessary because `self._vec.array` calls `VecGetArray/VecRestoreArray` which updates the state counter.

wence-: It seems like something is going wrong then because:
```
f = Function(V*V)
f.dat._vec.stateGet() => 1
f.dat._vec.array = 2
f.dat._vec.stateGet() => 2
f.dat._vec.array[1:3] = 3
f.dat._vec.stateGet() => 3
```

nbouziani: Ok thanks I will investigate that. Is the update of the state counter when `Vec._array` is called something recent ?

Increasing the state counter of the vec inside the `vec_context` is something that was originally in master so I assume there were a reason for that, e.g. handling a corner case.

nbouziani: Oh I just realized that we don't call `Vec._array` in WRITE mode so it won't increase the state for that. That doesn't answer the question about why `Vec._array` is not increasing the counter inside the `vec_context` but probably explains why the state counter was originally increased with `self._vec.stateIncrease()`.

wence-: > Oh I just realized that we don't call `Vec._array` in WRITE mode so it won't increase the state for that.

That is correct, because `access == WRITE` is a copy-out _from_ the single Vec _to_ the pieces (so it doesn't modify the single Vec.

nbouziani: Ok then I guess that means there were a problem with the initial code as well as the increase was also done for `access == WRITE`. Also, I have got the following mfe:

```
f = Function(V * V)
f.dat._vec.stateGet() => 1
f.dat._vec.array[1:3] = 1
f.dat._vec.stateGet() => 2
array = f.dat._vec.array
array[1:3] = 2
f.dat._vec.stateGet() => 2
```

I am probably missing something obvious here but I would have expected that updating `array` updates the counter.

wence-: Ah ok, there's some magic going on, `f.dat._vec.array` interfaces with the python buffer protocol, when that buffer is obtained, it calls `VecGetArray` and when the buffer is released it calls `VecRestoreArray`. It is the latter that increments the state counter. But in your code snippet you have not yet released the buffer (you're still holding a reference to `array`); if you put `del array` before the final `stateGet` then I think things should work.

nbouziani: Ah ok thanks, it's clearer now. It indeed works and adding this to the `vec_context` fixes the failing tests. It is still unclear to me why the counter was initially incremented via `self._vec.stateIncrease()` but at least we have a clear solution now.

pyop2/types/dat.py
```diff
@@ -1021,6 +1021,8 @@ def vec_context(self, access):
                     size = v.local_size
                     array[offset:offset+size] = v.array_r[:]
                     offset += size
+            del array
+
```
wence-, line 1025: I would prefer to make this lifetime explicitly managed:
```
with self._vec.array as array:
    for d in self:
        with d.vec_ro as v:
             ....
yield self._vec
....
```

nbouziani: Ok thanks! I think it is rather: `with self._vec as array` since `self._vec.array` is the numpy array. I am not sure if the PETSc Vec context manager will call `Vec.array` behind the scenes but we the counter does increase which is what we are after.

wence-: ah yes, it does

PR #674, opened 2022-10-04
add ComposedMap

ksagiyam:


ReubenHill:
What's this for? I presume it could replace my manual effort at https://github.com/firedrakeproject/firedrake/blob/ee720d61991e7f04004847dce79844771fad6e2b/firedrake/interpolation.py#L460-L509?

ksagiyam:
I need this for submesh. I think you can use `op2.ComposedMap(map2, map1)` for `composed_map(map1, map2)` except for the real function case.

ReubenHill:
> 

Do you implement caching? Also why would the real function case not work?

ksagiyam:
I meant those lines in your code:
```
    if map2 is None:
        # Real function space case
        return None
```

This is not cached now like other types of maps.

dham (changes requested):
Please put an extruded test here, and swap out Reuben's code in Firedrake (which will get us a parallel test).

dham (approved):

ksagiyam:
Rebased. Thank you!

PR #680, opened 2022-11-17
Add halo freezing

connorjward:
This lets us avoid doing intermediate halo exchanges when doing things like incrementing into the output tensor in `firedrake.assemble`.


pyop2/types/dat.py
```diff
@@ -539,14 +540,34 @@ def __itruediv__(self, other):
...
+    def freeze_halo(self):
+        """Disable halo exchanges.
+
+        This is useful in cases where one is repeatedly writing to (and not reading from)
+        a :class:`Dat` as the intermediate updates can be skipped.
+
+        Note that the user is responsible for ensuring that the :class:`Dat` is in a
+        consistent state prior to freezing (by calling :func:`Dat.global_to_local_begin`
+        and :func:`Dat.global_to_local_end`) and also after unfreezing (with
+        :func:`Dat.local_to_global_begin` and :func:`Dat.local_to_global_end`).
```
dham, line 553: This sounds like it's screaming out for a context manager which does this setup and unsetup for you. I.e.:
```python3
with dat.frozen_halo():
     assemble(...)
     assemble(...)
     assemble(...)
```

connorjward: You're quite right. I'll try to get that working.

wence-: I would advocate a model:

```python
def freeze_with(dat):
    dat.global_to_local_begin(op2.INC)
    dat.global_to_local_end(op2.INC)

def unfreeze_with(dat):
   ...

with dat.frozen_halo(pre=freeze_with, post=unfreeze_with):
    ...
```

Just FYI, don't use `contextlib.contextmanager` to make your context manager (since it is quite a bit slower than writing a class that provides a context manager interface by hand).

connorjward: I wish I had read this 20 minutes ago...

pyop2/types/dat.py
```diff
@@ -592,11 +593,58 @@ def local_to_global_end(self, insert_mode):
...
+                "results. Freezing halos is only valid for write-only access modes."
+            )
+
+        # First initialise the halo values (e.g. set to zero if INC'ing)
+        self.global_to_local_begin(access_mode)
+        self.global_to_local_end(access_mode)
+        self.freeze_halo()
+
+        # Execute main body of context manager
+        yield self
```
wence-, line 623: Another reason not to use `contextlib.contextmanager`: you _still_ need to manage exception handling in the body of your function. If executing the body raises an error, you won't run the cleanup code, which is kind of unexpected:

```python
In [7]: @contextlib.contextmanager
   ...: def foo():
   ...:     print("before")
   ...:     yield
   ...:     print("after")
   ...: 

In [8]: with foo():
   ...:     print("body")
   ...: 
before
body
after

In [9]: with foo():
   ...:     print("body")
   ...:     raise RuntimeError
   ...: 
before
body
RuntimeError                              Traceback (most recent call last)
Input In [9], in <cell line: 1>()
      1 with foo():
      2     print("body")
----> 3     raise RuntimeError

RuntimeError: 
```

pyop2/types/dat.py
connorjward: Arguably the performance difference is not worth the effort of writing something more complicated. The context manager is supposed to live outside parloops so there's no way this is ever a bottleneck.

pyop2/types/dat.py
wence-: No, just remember to do it if you want the cleanup to occur!

pyop2/types/dat.py
wence-: FWIW, the "more complicated thing" is:
```python
class frozen_halo:
    def __init__(self, pre, post):
        self.pre = pre
        self.post = post
    def __enter__(self):
         self.pre()
    def __exit__(self, exc_typ, exc_val, tb):
         # handle exception if you like
         self.post()

class Dat:
     def frozen_halo(self, access):
          return frozen_halo(...)
```

pyop2/types/dat.py
wence-: So:
```python
try:
    yield self
finally:
    self.unfreeze_halo()
    self.local_to_global(..)
    ...
```

connorjward: Sure. I was confused because I don't think of halo exchanges as 'cleanup' necessarily.

wence-: I guess if something fails in the parloop body there's probably not a lot you can end up doing, but maybe you could imagine catching some things and you would want a consistent state for your halo afterwards.

pyop2/types/dat.py
connorjward: Thanks. I'll have a go at implementing this.

dham (approved):

PR #682, opened 2022-11-21
Add timeout to CI tests

JDBetteridge:
Now we have a matrix of tests we should take steps to avoid burning CI hours when tests are hanging or we introduce severe performance regressions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,11 +13,11 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
-    # strategy:
-    #   # Don't immediately kill all if one Python version fails
-    #   fail-fast: false
-    #   matrix:
-    #     python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
+    strategy:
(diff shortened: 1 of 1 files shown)


JDBetteridge:
Found the error (check previous CI runs)
```
A process has executed an operation involving a call
to the fork() system call to create a child process.

As a result, the libfabric EFA provider is operating in
a condition that could result in memory corruption or
other system errors.

For the libfabric EFA provider to work safely when fork()
is called, you will need to set the following environment
variable:
          RDMAV_FORK_SAFE

However, setting this environment variable can result in
signficant performance impact to your application due to
increased cost of memory registration.

You may want to check with your application vendor to see
if an application-level alternative (of not using fork)
exists.

Your job will now abort.
```

Fixed by setting `RDMAV_FORK_SAFE=1` which may incur a slight performance penalty on CI for Python 3.10/3.11, which were previously passing.

connorjward (approved):
LGTM

JDBetteridge:
An alternative to this is to use `PYOP2_NO_FORK_AVAILABLE=1`, but this is currently broken (see #683)

PR #688, opened 2023-01-27
Test updated types without __getattr__

Straggio:


pyop2/types/dataset.py
```diff
@@ -330,32 +335,22 @@ def dm(self):
 
 class MixedDataSet(DataSet):
     r"""A container for a bag of :class:`DataSet`\s.
-
     Initialized either from a :class:`MixedSet` and an iterable or iterator of
     ``dims`` of corresponding length ::
-
```
connorjward, line 336: Why have these been modified? I think the blank lines here aid readability to distinguish the code blocks from the rest of the text.

pyop2/types/set.py
```diff
@@ -345,10 +333,17 @@ def _argtypes_(self):
     def _wrapper_cache_key_(self):
         return self.parent._wrapper_cache_key_ + (self.constant_layers, )
 
-    def __getattr__(self, name):
-        """Returns a :class:`Set` specific attribute."""
-        value = getattr(self._parent, name)
-        return value
+    @utils.cached_property
```
connorjward, line 336: I don't think that we need `cached_property` for pass-through properties like this. If this is an expensive thing to compute then we should cache it for `self._parent.sizes`, not here.

pyop2/types/set.py
```diff
@@ -424,11 +418,13 @@ def _kernel_args_(self):
...
         return self._superset._argtypes_ + (ctypes.c_voidp, )
 
-    # Look up any unspecified attributes on the _set.
-    def __getattr__(self, name):
-        """Returns a :class:`Set` specific attribute."""
-        value = getattr(self._superset, name)
-        return value
+    @property
+    def _parent(self):
+        return self._superset
```
connorjward, line 423: Shouldn't this be `self._superset._parent`?

pyop2/types/dataset.py
```diff
@@ -244,7 +251,7 @@ def set(self):
     @utils.cached_property
     def size(self):
         """The number of local entries in the Dataset (1 on rank 0)"""
-        return 1 if mpi.MPI.comm.rank == 0 else 0
+        return 1 if self.comm.rank == 0 else 0
```
connorjward, line 254: This should be put in a distinct commit as it is unrelated to the rest of the changes here.

connorjward (changes requested):
Some initial feedback. Thanks for working on this.

pyop2/types/dataset.py
Straggio: Seems like my editor just deleted them. I will revert it.

pyop2/types/set.py
Straggio: No, because the superset could also be a Set and the class `Set` don't have the attribute _parent. But I looked into it and `_parent` is used to get the attribute `name` in a specific call `self._parent.name`maybe inherited from the `ExtrudedSet` but not sure. I can change it to 
```
@property
 def name(self):
     return self._superset.name
```
I think this makes more sense. We then need to run the CI again to make sure there isn't another attribute called over this `_parent` attribute.

pyop2/types/set.py
Straggio: Here I am not sure. I couldn't find a way to make it work with just `@property`. The error message looks like this:
```bash
__________________________________________________________ TestSubsetAPI.test_empty_subset ___________________________________________________________

self = <test_api.TestSubsetAPI object at 0x7fc13b2facb0>, set = Set((5, 5, 5), 'foo')

    def test_empty_subset(self, set):
        "Subsets can be empty."
>       ss = op2.Subset(set, [])

test/unit/test_api.py:328: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
<decorator-gen-21>:2: in __init__
    ???
pyop2/utils.py:133: in wrapper
    return f(*args, **kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = Subset(Set((5, 5, 5), 'foo'), array([], dtype=int32)), superset = Set((5, 5, 5), 'foo'), indices = array([], dtype=float64)

    @utils.validate_type(('superset', Set, TypeError),
                         ('indices', (list, tuple, np.ndarray), TypeError))
    def __init__(self, superset, indices):
        self.comm = mpi.internal_comm(superset.comm)
    
        # sort and remove duplicates
        indices = np.unique(indices)
        if isinstance(superset, Subset):
            # Unroll indices to point to those in the parent
            indices = superset.indices[indices]
            superset = superset.superset
        assert type(superset) is Set or type(superset) is ExtrudedSet, \
            'Subset construction failed, should not happen'
    
        self._superset = superset
        self._indices = utils.verify_reshape(indices, dtypes.IntType, (len(indices),))
    
        if len(self._indices) > 0 and (self._indices[0] < 0 or self._indices[-1] >= self._superset.total_size):
            raise ex.SubsetIndexOutOfBounds(
                'Out of bounds indices in Subset construction: [%d, %d) not [0, %d)' %
                (self._indices[0], self._indices[-1], self._superset.total_size))
    
>       self._sizes = ((self._indices < superset.core_size).sum(),
                       (self._indices < superset.size).sum(),
                       len(self._indices))
E       AttributeError: can't set attribute '_sizes'

pyop2/types/set.py:418: AttributeError
```
Since the problem occured in utils.py I changed it to `@utils.cached_property` and it worked.

connorjward: Ah. I think we are running into one of the other big design issues with PyOP2 which is that the inheritance is super confusing. I don't think we need to be too fussy here.

pyop2/types/set.py
connorjward: Yeah that makes more sense. Go for it.

pyop2/types/set.py
connorjward: So to be clear, you can leave it like this.

pyop2/types/set.py
Straggio: You can now run the Firedrake CI once more.

PR #698, opened 2023-05-02
DO NOT MERGE Check that upstream changes have not broken CI

connorjward:


connorjward:
See https://github.com/OP2/PyOP2/pull/700

Issue #693, opened 2023-03-21
Add global cache manager to facilitate cleaning memory


connorjward:
In lots of different places in PyOP2 and Firedrake we have global caches implemented something like:

```py
class MyObject:
    _cache = {}
```

For long running simulations (e.g. CI) this results in a slow leaking of memory as this global cache gets larger and larger. Thetis for example needs to [purge](https://github.com/thetisproject/thetis/blob/master/test/conftest.py#L10) the cache at every test to avoid running out of memory.

It would be nice to have some sort of global `pyop2.caching.cache_manager` object to hide this. This would mean class-level caches would be implemented something like:

```py
class MyObject:
    _cache = cache_manager.register("MyObject")
```

then later on in the simulation all (or some) of these caches can be cleaned using `cache_manager.clear()`.

connorjward:
Closing as this is sorted in https://github.com/OP2/PyOP2/pull/724

PR #703, opened 2023-06-12
Add parallel-safe LRU cache and global cache manager objects

### LINKED ISSUE #696: Add LRU cache
We should introduce an LRU cache object/function to prevent unbounded memory usage by our in-memory caches. We will need to be careful about including the communicator in the cache key somehow so we never evict items from some ranks but not others.

Related issue #693.
Related discussion https://github.com/firedrakeproject/firedrake/discussions/2865

connorjward:
There are occasions in PyOP2/Firedrake where we repeatedly perform expensive operations but are unable to cache them because:

- No suitable object exists upon which the object may be cached, and
- The object references "large" data structures

A classic example of this is something like computing the [`adjoint`](https://github.com/firedrakeproject/firedrake/blob/c5cd37ad718da017f3035a5a7c6f5329d60e62aa/firedrake/ufl_expr.py#L214) of a form. Forms are often non-persistent (e.g. `solve(u*v*dx == f*v*dx, ...)`) and they hold references to the coefficients so a global cache is not suitable. Computing `adjoint` is a significant cost when computing the derivative of adjoint problems so caching results is desirable.

This PR introduces parallel-safe LRU caches. I believe that this would address the problems described above.

I have also introduced the concept of a "cache manager" (https://github.com/OP2/PyOP2/issues/693) to facilitate cache cleaning and inspection.

Closes #696 #693 

TODOs:

- [ ] Code cleanup and documentation
- [ ] Write tests
- [ ] Enable configuration of cache sizes via environment variables - users with memory constraints would want smaller caches

### DIFF
--- a/pyop2/caching.py
+++ b/pyop2/caching.py
@@ -349,6 +349,8 @@ def _disk_cache_set(cachedir, key, value):
     tempfile.rename(filepath)
 
 
+# TODO LRU caches should probably emit a warning when they overflow, would make
+# development easier
 class CountedCache(collections.abc.MutableMapping, abc.ABC):
     def __init__(self):
         self.naccesses = 0
--- a/pyop2/caching.py
+++ b/pyop2/caching.py
@@ -33,15 +33,19 @@
 
 """Provides common base classes for cached objects."""
 
+import abc
+import collections
 import hashlib
 import os
-from pathlib import Path
 import pickle
+from collections import defaultdict
+from functools import partial
+from pathlib import Path
 
 import cachetools
 
 from pyop2.configuration import configuration
-from pyop2.mpi import hash_comm
+from pyop2.mpi import COMM_WORLD, hash_comm, temp_internal_comm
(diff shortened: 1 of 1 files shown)


pyop2/caching.py
```diff
@@ -343,3 +347,147 @@ def _disk_cache_set(cachedir, key, value):
...
+    """Parallel-safe LRU cache."""
+
+    def __init__(self, maxsize=32):
+        super().__init__()
+        self._caches = defaultdict(partial(CountedLRUCache, maxsize))
+        self.maxsize = maxsize
+
+
+
+class CacheManager(dict):
```
connorjward, line 463: @JDBetteridge: we should have pretty printing of cache statistics.

connorjward:
Closing as @JDBetteridge has [already merged](https://github.com/OP2/PyOP2/pull/724) a much nicer version of what I was trying to accomplish.

PR #708, opened 2023-09-21
Passthrough params

pbrubeck:


connorjward (commented):
I am happy with this but I wrote all the code! We can review this on Wednesday.

connorjward (approved):
I reckon this is fine. It is a hack and not super well documented, but it should be straightforward (and cleaner) to make pyop3 handle this in the near future.

dham (approved):

PR #710, opened 2023-10-17
Fix halo exchanges for MixedDats in parloops

connorjward:
Fixes https://github.com/firedrakeproject/firedrake/issues/3150

Very annoyingly I have had to add a test for this to Firedrake since we don't have the right parallel infrastructure in PyOP2. I want to move it back when pyop3 lands.

### DIFF
--- a/pyop2/global_kernel.py
+++ b/pyop2/global_kernel.py
@@ -6,8 +6,9 @@ import os
 from typing import Optional, Tuple
 
 import loopy as lp
-from petsc4py import PETSc
 import numpy as np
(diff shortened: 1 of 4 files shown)


ksagiyam (changes requested):
Could you also fix the friend at line 366 in `parloop.py`?

connorjward:
> Could you also fix the friend at line 366 in `parloop.py`?

Done. Thanks for spotting that!

ksagiyam (approved):

PR #711, opened 2023-11-14
Remove __del__ method and add weakref.finalizer

JDBetteridge:
This PR changes the interface of `pyop2.mpi.internal_comm` to include an additional argument. This is the object to which the comm is associated, and the lifetime of which the internal comm (reference) is associated with.

Requires #712 to go in first

### DIFF
--- a/pyop2/caching.py
+++ b/pyop2/caching.py
@@ -83,13 +83,6 @@ class ObjectCached(object):
     details).  The object on which the cache is stored should contain
     a dict in its ``_cache`` attribute.
 
-    .. warning ::
-
-       This kind of cache sets up a circular reference.  If either of
-       the objects implements ``__del__``, the Python garbage
-       collector will not be able to collect this cycle, and hence
-       the cache will never be evicted.
-
     .. warning::
 
         The derived class' :meth:`__init__` is still called if the
--- a/pyop2/mpi.py
+++ b/pyop2/mpi.py
@@ -42,6 +42,7 @@ import gc
 import glob
 import os
 import tempfile
+import weakref
 
 from pyop2.configuration import configuration
 from pyop2.exceptions import CompilationError
@@ -267,9 +268,7 @@ class temp_internal_comm:
     def __init__(self, comm):
         self.user_comm = comm
         self.internal_comm = internal_comm(self.user_comm)
-
-    def __del__(self):
-        decref(self.internal_comm)
+        weakref.finalize(self, decref, self.internal_comm)
 
     def __enter__(self):
         """ Returns an internal comm that will be safely decref'd
--- a/pyop2/mpi.py
+++ b/pyop2/mpi.py
@@ -267,8 +267,7 @@ class temp_internal_comm:
     """
     def __init__(self, comm):
         self.user_comm = comm
-        self.internal_comm = internal_comm(self.user_comm)
-        weakref.finalize(self, decref, self.internal_comm)
+        self.internal_comm = internal_comm(self.user_comm, self)
 
     def __enter__(self):
         """ Returns an internal comm that will be safely decref'd
@@ -282,10 +281,12 @@ class temp_internal_comm:
         pass
 
 
-def internal_comm(comm):
+def internal_comm(comm, obj):
     """ Creates an internal comm from the user comm.
     If comm is None, create an internal communicator from COMM_WORLD
     :arg comm: A communicator or None
+    :arg obj: The object which the comm is an attribute of
+    (usually `self`)
 
     :returns pyop2_comm: A PyOP2 internal communicator
     """
@@ -308,6 +309,7 @@ def internal_comm(comm):
         pyop2_comm = comm
     else:
         pyop2_comm = dup_comm(comm)
+    weakref.finalize(obj, decref, pyop2_comm)
     return pyop2_comm
 
 
@@ -444,10 +446,13 @@ def set_compilation_comm(comm, comp_comm):
 
 
 @collective
-def compilation_comm(comm):
+def compilation_comm(comm, obj):
     """Get a communicator for compilation.
 
     :arg comm: The input communicator, must be a PyOP2 comm.
+    :arg obj: The object which the comm is an attribute of
+    (usually `self`)
+
     :returns: A communicator used for compilation (may be smaller)
     """
     if not is_pyop2_comm(comm):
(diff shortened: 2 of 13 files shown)


pyop2/caching.py
```diff
@@ -83,13 +83,6 @@ class ObjectCached(object):
     details).  The object on which the cache is stored should contain
     a dict in its ``_cache`` attribute.
 
-    .. warning ::
-
-       This kind of cache sets up a circular reference.  If either of
-       the objects implements ``__del__``, the Python garbage
-       collector will not be able to collect this cycle, and hence
-       the cache will never be evicted.
-
```
JDBetteridge, line 92: https://docs.python.org/3/library/gc.html#gc.garbage

pyop2/mpi.py
```diff
@@ -272,10 +281,12 @@ def __exit__(self, exc_type, exc_value, traceback):
         pass
 
 
-def internal_comm(comm):
+def internal_comm(comm, obj):
     """ Creates an internal comm from the user comm.
     If comm is None, create an internal communicator from COMM_WORLD
     :arg comm: A communicator or None
+    :arg obj: The object which the comm is an attribute of
```
connorjward, line 288: I don't love this pattern. Would something like a mixin class be better?

pyop2/mpi.py
```diff
@@ -465,26 +474,41 @@ def compilation_comm(comm):
     else:
         comp_comm = comm
     incref(comp_comm)
+    weakref.finalize(obj, decref, comp_comm)
     return comp_comm
 
 
+def finalize_safe_debug():
```
connorjward, line 481: Needs a docstring. Very unclear what this is doing.

pyop2/mpi.py
```diff
@@ -465,26 +474,41 @@ def compilation_comm(comm):
...
+    if PYOP2_FINALIZED:
+        if logger.level > DEBUG and not _running_on_ci:
+            debug = lambda string: None
+        else:
+            debug = lambda string: print(string)
+    return debug
```
connorjward, line 487: I assume this is a global variable? Otherwise this will break.

pyop2/mpi.py
```diff
@@ -465,26 +474,41 @@ def compilation_comm(comm):
...
-        debug = lambda string: None
-    else:
-        debug = lambda string: print(string)
+    debug = finalize_safe_debug()
     debug("PyOP2 Finalizing")
     # Collect garbage as it may hold on to communicator references
+
     debug("Calling gc.collect()")
     gc.collect()
+    debug("STATE0")
```
connorjward, line 501: Not a very helpful statement to have here.

pyop2/types/mat.py
```diff
@@ -383,10 +381,10 @@ def __init__(self, parent, i, j):
         self._dims = tuple([tuple([parent.dims[i][j]])])
         self._blocks = [[self]]
         self.iteration_regions = parent.iteration_regions
-        self.lcomm = mpi.internal_comm(self.dsets[0].comm)
-        self.rcomm = mpi.internal_comm(self.dsets[1].comm)
+        self.lcomm = mpi.internal_comm(self.dsets[0].comm, self)
```
connorjward, line 384: Is it necessary to increment the reference count for the internal comm here? Could we not have something like:
```py
@property
def lcomm(self):
    return self.dsets[0]._comm
```

pyop2/types/set.py
```diff
@@ -543,13 +537,15 @@ def __init__(self, sets):
         assert all(s is None or isinstance(s, GlobalSet) or ((s.layers == self._sets[0].layers).all() if s.layers is not None else True) for s in sets), \
             "All components of a MixedSet must have the same number of layers."
         # TODO: do all sets need the same communicator?
-        self.comm = mpi.internal_comm(functools.reduce(lambda a, b: a or b, map(lambda s: s if s is None else s.comm, sets)))
+        self.comm = mpi.internal_comm(
+            functools.reduce(
+                lambda a, b: a or b,
+                map(lambda s: s if s is None else s.comm, sets)
+            ),
```
connorjward, line 544:
```suggestion
            pytools.single_valued(s.comm for s in sets if s is not None)
```

pyop2/mpi.py
JDBetteridge: Yes, it's global

pyop2/mpi.py
JDBetteridge: Do you have a suggestion, I was only using `STATE {0,1,2}` for debugging and was potentially going to remove the "titles"

pyop2/types/mat.py
JDBetteridge: No, but it's easier to use the pattern than to break it in this one instance. The whole `lcomm` and `rcomm` thing seems redundant to me, it isn't actually being used. I think the original code was aspirational and the functionality that would utilise it never implemented in PyOP2. There's a comment to that effect somewhere in the code.

pyop2/types/set.py
JDBetteridge: Do I need to add pytools as a hard dependency to do this?

pyop2/mpi.py
JDBetteridge: Not my favourite either. I agree this should derive from an `MPIObject` base class, but the place for that to live is either `mpi4py` or a dedicated add on library and this is a future task. I'd rather not refactor all the PyOP2 objects (again).

pyop2/types/mat.py
wence-: Yeah, petsc operators don't have different left and right communicators so this is over-eager and could be culled and replaced with a check in the constructor that the communicators of the row and column sets are the same.

pyop2/types/set.py
wence-: Or `val, = set(s.comm for s in sets if s is not None)` if you don't want pytools.

JDBetteridge: Nice idea, but `mpi4py.MPI.Intracomm` is not a hashable type, so using a Python `set()` won't work here.

connorjward: We always get pytools with loopy, and writing a replacement would be trivial if we eventually rip loopy out, so I see no issue in using it here. It should be listed as a formal requirement though.

JDBetteridge: Yes, this is done :slightly_smiling_face:

JDBetteridge:
I'm merging this into [PR 712](https://github.com/OP2/PyOP2/pull/712) to make my life a bit easier.

PR #720, opened 2024-04-19
Fix for massively parallel performance regression

### LINKED ISSUE #719: `PYOP2_NODE_LOCAL_COMPILATION=0` invokes compiler on non-root ranks
It's my understanding that the default for the `node_location_configuration` is True, i.e. all ranks will invoke the compiler inside `pyop2.compilation.Compiler.get_so()`. However, when this setting is False, only rank 0 will invoke the compiler, and the assumption is that all other ranks are able to access the outputs from this. (As an aside, this doesn't seem to capture the scenario of running on only one rank of a physical node when running across nodes with MPI?)

For context, we're running Firedrake with a fairly large number of ranks, and there's a significant disk impact when all these ranks simultaneously invoke the compiler for version sniffing. As far as I can tell, there are two places this happens:
1. Within `pyop2.compilation.sniff_compiler()` itself, as called from `pyop2.compilation.load()`. In this case, the non-compiling ranks can probably just use the `AnonymousCompiler` or `Compiler` base class: they just need to wait for the MPI barrier to signal the shared library is available, and load it. I think in this case, the defaults are such that `Compiler.sniff_compiler_version()` will simply run into an exception because no compiler is defined.
2. If, e.g. `pyop2.compilation.set_default_compiler(pyop2.compilation.LinuxGnuCompiler)` is called to avoid the above call to `sniff_compiler()`, the compiler is still invoked on all ranks: `Compiler.__init__()` calls `Compiler.sniff_compiler_version()`, which will now have a valid executable.

As far as I can tell, `Compiler.sniff_compiler_version()` is only required for cflags patching for certain (very old) compiler versions. In order not to incur the wrath of our HPC sysadmins, I've resorted to the following preamble to patch out all the sniffing behaviour:

```python
import pyop2.compilation

class C(pyop2.compilation.LinuxGnuCompiler):
    def sniff_compiler_version(self, *args): pass

    @property
    def bugfix_cflags(self): return ()

pyop2.compilation.set_default_compiler(C)
```

However I think the core issue is that, in my opinion, the compiler simply doesn't need to be run on non-compiling ranks.

JDBetteridge:
Fixes #719

### DIFF
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -249,7 +249,7 @@ class Compiler(ABC):
         # MPI ranks is expensive
         exe = self.cxx if cpp else self.cc
         version = None
-        if mpi.COMM_WORLD.rank == 0:
+        if self.comm.rank == 0:
             # `-dumpversion` is not sufficient to get the whole version string (for some compilers),
             # but other compilers do not implement `-dumpfullversion`!
             for dumpstring in ["-dumpfullversion", "-dumpversion"]:
@@ -265,7 +265,7 @@ class Compiler(ABC):
                     break
                 except (subprocess.CalledProcessError, UnicodeDecodeError, InvalidVersion):
                     continue
-        self.version = mpi.COMM_WORLD.bcast(version, 0)
+        self.version = self.comm.bcast(version, 0)
 
     @property
     def bugfix_cflags(self):
--- a/pyop2/compilation.py
+++ b/pyop2/compilation.py
@@ -89,12 +89,14 @@ def sniff_compiler(exe, comm=mpi.COMM_WORLD):
     """Obtain the correct compiler class by calling the compiler executable.
 
     :arg exe: String with name or path to compiler executable
+    :arg comm: Comm over which we want to determine the compiler type
     :returns: A compiler class
     """
-    # Note:
-    # Sniffing compiler for very large numbers of MPI ranks is expensive
     compiler = None
     if comm.rank == 0:
+        # Note:
+        # Sniffing compiler for very large numbers of MPI ranks is
+        # expensive so we do this on one rank and broadcast
         try:
             output = subprocess.run(
                 [exe, "--version"],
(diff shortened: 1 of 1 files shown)


pyop2/compilation.py
```diff
@@ -85,66 +85,71 @@ def set_default_compiler(compiler):
         )
 
 
-def sniff_compiler(exe):
+def sniff_compiler(exe, comm=mpi.COMM_WORLD):
     """Obtain the correct compiler class by calling the compiler executable.
 
     :arg exe: String with name or path to compiler executable
```
connorjward, line 91: Missing `comm` argument.

pyop2/compilation.py
```diff
@@ -85,66 +85,71 @@ def set_default_compiler(compiler):
...
-        if name == "Intel":
-            compiler = LinuxIntelCompiler
-        elif name == "GNU":
-            compiler = LinuxGnuCompiler
-        elif name == "clang":
-            compiler = LinuxClangCompiler
-        elif name == "Cray":
-            compiler = LinuxCrayCompiler
+    # Note:
+    # Sniffing compiler for very large numbers of MPI ranks is expensive
```
connorjward, line 95: Probably put this in the docstring rather than here.

pyop2/compilation.py
```diff
@@ -178,8 +183,8 @@ class Compiler(ABC):
     _debugflags = ()
 
     def __init__(self, extra_compiler_flags=(), extra_linker_flags=(), cpp=False, comm=None):
-        # Get compiler version ASAP since it is used in __repr__
-        self.sniff_compiler_version()
+        # Set compiler version ASAP since it is used in __repr__
+        self.version = None
```
connorjward, line 187: This is definitely a code smell, but I don't think it's worth worrying about too much.

To address it I would personally write
```py
@cached_property
def version(self):
    # sniff compiler version...
```
so nothing then needs to be done in `__init__`.

pyop2/compilation.py
```diff
@@ -552,7 +546,7 @@ class AnonymousCompiler(Compiler):
 
 @mpi.collective
 def load(jitmodule, extension, fn_name, cppargs=(), ldargs=(),
-         argtypes=None, restype=None, comm=None):
+         argtypes=None, restype=None, comm=mpi.COMM_WORLD):
```
connorjward, line 549: Is `comm` getting passed correctly into this function by PyOP2?

connorjward (dismissed):
Just some small tweaks. When it goes in I'll have to port this to pyop3 :/

pyop2/compilation.py
JDBetteridge: I will revert this change. Originally I was doing communication in this routine

dham (approved):

PR #724, opened 2024-08-08
Remove comm hash and add per-comm caches

JDBetteridge:
It's broken and it's dangerous!

Now that we actually free our duplicated comms the `hash_comm` function is less than useless as the same `id` can be returned by different comms.

This PR also adds per-comm caches, which circumvents the need to hash a communicator in the first place.

### Update:
This (necessarily) turned into a much bigger refactor than was intended. The key changes, beyond removing comm hash, are:
- Removal of `Cached` a base class used to create an object that would be cached. However, overriding the `__new__` method effectively turned instances into singletons, but the objects didn't necessarily fit this design pattern. Moving forward only functions should be wrapped in a cache.
- Removal of `@cached_property` decorator as this is now part of the Python standard library.
- Refactor of `load` and `get_so` in `compilation.py` to fit with the new caching framework. This now compiles shared objects to RAM backed storage (`/tmp`) and only caches the `.so` to non-volatile storage (`$PYOP2_CACHE_DIR`).
- Add a `PYOP2_SPMD_STRICT` environment variable, for checking SPMD behaviour by enabling `MPI.Comm.Barrier()` on logically collective calls. (Switched off by default)

Moving forward (in pyop3???) we should also remove the `ObjectCached` base class as it suffers from the same issues as `Cached`, but this is another big refactor.


pyop2/caching.py
```diff
@@ -240,14 +241,109 @@ def cache_key(self):
...
     may evict results leading to deadlocks.
 """
 
 
+def default_parallel_hashkey(*args, **kwargs):
+    comm = kwargs.get('comm')
+    return comm, cachetools.keys.hashkey(*args, **kwargs)
+
+
+def parallel_memory_only_cache(key=default_parallel_hashkey):
```
connorjward, line 256: Bit of a mouthful, `parallel_cache` instead? then could have `parallel_disk_cache` if needed.

pyop2/caching.py
```diff
@@ -240,14 +241,109 @@ def cache_key(self):
...
+            return local_cache.setdefault(k, v)
+
+        return wrapper
+    return decorator
+
+
+def parallel_memory_only_cache_no_broadcast(key=default_parallel_hashkey):
+    """Memory only cache decorator.
+
+    Decorator for wrapping a function to be called over a communiucator in a
```
connorjward, line 305:
```suggestion
    Decorator for wrapping a function to be called over a communicator in a
```

pyop2/caching.py
```diff
@@ -240,14 +241,109 @@ def cache_key(self):
...
+        communicator to be collective over and the second is the key. This is
+        required to ensure that deadlocks do not occur when using different
+        subcommunicators.
+    """
+    def decorator(func):
+        def wrapper(*args, **kwargs):
+            """ Extract the key and then try the memory cache before falling back
+            on calling the function and populating the cache.
+            """
+            comm, mem_key = key(*args, **kwargs)
```
connorjward, line 274: Maybe it would be "cleaner" to have 2 callables, one returning the comm and the other the hash key. The 2-tuple approach was needed for the `collective` kwarg to work I think.

connorjward (commented):
procrastinating

.github/workflows/ci.yml
```diff
@@ -28,6 +28,13 @@ jobs:
     timeout-minutes: 60
 
     steps:
+      - name: Setup tmate session
+        uses: mxschmitt/action-tmate@v3
+        with:
+          detached: true
+          limit-access-to-actor: true
+        timeout-minutes: 15
+
```
JDBetteridge, line 37:
```suggestion
```
Turns out this isn't safe to merge to master (it's fine to use it in branches though)

pyop2/caching.py
```diff
@@ -160,179 +149,427 @@ def make_obj():
...
+        :arg value: The new item to store in the cache.
+        """
+        k1, k2 = key[0][:2], key[0][2:] + key[1]
+        basedir = Path(self.cachedir, k1)
+        basedir.mkdir(parents=True, exist_ok=True)
 
+        tempfile = basedir.joinpath(f"{k2}_p{os.getpid()}.tmp")
+        filepath = basedir.joinpath(k2)
+        with self.open(tempfile, mode="wb") as fh:
+            self.write(fh, value)
```
dham, line 325: Please use [mkstemp()](https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp)

pyop2/compilation.py
```diff
@@ -604,6 +489,179 @@ def __init__(self, code, argtypes):
...
+    :arg extension: extension of the source file (c, cpp).
+    :arg comm: Communicator over which to perform compilation.
+    :arg filename: Optional
+    Returns a :class:`ctypes.CDLL` object of the resulting shared
+    library."""
+    if filename is None:
+        # A UUID should ensure we have a unique path
+        uuid = uuid4().hex
+        # Taking the first two characters avoids using excessive filesystem inodes
+        tempdir = MEM_TMP_DIR.joinpath(f"{uuid[:2]}")
```
dham, line 582: do this the official way,

pyop2/compilation.py
```diff
@@ -604,6 +489,179 @@ def __init__(self, code, argtypes):
...
+        compiler_flags = compiler.cflags
+
+    # TODO: Do we still need to worry about atomic file renaming in this function?
+    base = filename.stem
+    path = filename.parent
+    pid = os.getpid()
+    cname = filename.with_name(f"{base}_p{pid}.{extension}")
+    oname = filename.with_name(f"{base}_p{pid}.o")
+    # Link into temporary file, then rename to shared library atomically (avoiding races).
+    tempname = filename.with_stem(f"{base}_p{pid}.so")
```
dham, line 608: filenames

pyop2/compilation.py
```diff
@@ -184,23 +229,22 @@ class Compiler(ABC):
...
+        self._debug = debug
 
     def __repr__(self):
-        return f"<{self._name} compiler, version {self.version or 'unknown'}>"
+        string = f"{self.__class__.__name__}("
+        string += f"extra_compiler_flags={self._extra_compiler_flags}, "
+        string += f"extra_linker_flags={self._extra_linker_flags}, "
+        string += f"version={self._version!r}, "
+        string += f"debug={self._debug})"
+        return string
```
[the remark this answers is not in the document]
JDBetteridge: Noted! (But not relevant in this instance and I think this looks better than one of the more efficient alternatives)

connorjward (approved):
Looks fine.

PR #727, opened 2024-08-29
DatView: Fix zero()

pbrubeck:
Fix `DatView.zero()` by overriding `DatView._data` to return the parent slice.


connorjward (changes requested):
Please add a test and make sure this doesn't break Firedrake CI.

pyop2/types/dat.py
```diff
@@ -722,39 +723,11 @@ def halo_valid(self, value):
...
-        idx = (slice(None), *self.index)
-        return full[idx]
-
-    @property
-    def data_wo_with_halos(self):
-        full = self._parent.data_wo_with_halos
-        idx = (slice(None), *self.index)
-        return full[idx]
+    def _data(self):
+        return self._parent._data[self._idx]
```
wence-, line 730: Will halo updates still do the right thing after this change and mark the parent's halos as dirty?

pbrubeck: It should, as the halo_valid property is mapped to that of the parent https://github.com/OP2/PyOP2/blob/5316660eb1ef9a84a1484293bb23083a85b2b60d/pyop2/types/dat.py#L716-L722

dham (approved):

